PR #8, opened 2022-10-04
fix: correctly grow memory on values of size 0

frederikrothenberger:
The cell implementation did not correctly grow the memory to store its header when initialized with a default value of size 0.


ielashi (approved):
LGTM!

PR #17, opened 2022-11-16
chore: Enhancements to the StableBTreeMap API.

ielashi:
BTreeMap currently take the maximum size of the BTreeMap as arguments. Example:

```
let map: StableBTreeMap<_, Key, Value> = StableBTreeMap::init(memory, MAX_KEY_SIZE, MAX_VALUE_SIZE)
```

One issue with this API is that specifying the max sizes is independent of the type being used, and that poses a couple of usability issues:

1.  If you're using a primitive type as your key/value (e.g. `u64`), it's not obvious what size you should put as an argument. The correct answer is `8`, because the `Storable` implementation of `u64` serializes it to a vector of 8 bytes, but that's far from obvious.
2. It's easy to put an incorrect value by accident if you're managing a few stable structures. I already made this mistake in the bitcoin canister.

A more safe and clear implementation is that the `Key` and `Value` contain information about their size, so that users don't need to specify the size explicitly when initializing the map.

This commit introduces a `BoundedStorable` trait, where it now required for a type to be used in `StableBTreeMap`. The map is then able to fetch the maximum size of that type without it being passed explicitly.

```
impl BoundedStorable for Key {
  fn max_size() -> u32 {
     MAX_KEY_SIZE
  }
}

impl BoundedStorable for Value {
  fn max_size() -> u32 {
     MAX_VALUE_SIZE
  }
}
```

And now the syntax is closer to a standard `BTreeMap`:

```
let map: StableBTreeMap<_, Key, Value> = StableBTreeMap::init(memory)
```

Storing the maximum size in the type allows us in the future to also support blanket implementations of tuples, so one can then implement:

```
let map: StableBTreeMap<_, (A, B), (C, D)> = StableBTreeMap::init(memory)
```

The updated examples showcase how the new API is used.

### DIFF
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -5,7 +5,7 @@ mod node;
 use crate::{
     read_struct,
     types::{Address, Bytes, NULL},
-    write_struct, Memory, Storable,
+    write_struct, BoundedStorable, Memory,
 };
 use allocator::Allocator;
 pub use iter::Iter;
@@ -20,7 +20,7 @@ const MAGIC: &[u8; 3] = b"BTR";
 ///
 /// The implementation is based on the algorithm outlined in "Introduction to Algorithms"
 /// by Cormen et al.
-pub struct BTreeMap<M: Memory, K: Storable, V: Storable> {
+pub struct BTreeMap<M: Memory, K: BoundedStorable, V: BoundedStorable> {
     // The address of the root node. If a root node doesn't exist, the address
     // is set to NULL.
     root_addr: Address,
@@ -61,15 +61,15 @@ impl BTreeHeader {
     }
 }
 
-impl<M: Memory + Clone, K: Storable, V: Storable> BTreeMap<M, K, V> {
+impl<M: Memory + Clone, K: BoundedStorable, V: BoundedStorable> BTreeMap<M, K, V> {
     /// Initializes a `BTreeMap`.
     ///
     /// If the memory provided already contains a `BTreeMap`, then that
     /// map is loaded. Otherwise, a new `BTreeMap` instance is created.
-    pub fn init(memory: M, max_key_size: u32, max_value_size: u32) -> Self {
+    pub fn init(memory: M) -> Self {
         if memory.size() == 0 {
             // Memory is empty. Create a new map.
-            return BTreeMap::new(memory, max_key_size, max_value_size);
+            return BTreeMap::new(memory);
         }
 
         // Check if the magic in the memory corresponds to a BTreeMap.
@@ -77,7 +77,7 @@ impl<M: Memory + Clone, K: Storable, V: Storable> BTreeMap<M, K, V> {
         memory.read(0, &mut dst);
         if dst != MAGIC {
             // No BTreeMap found. Create a new instance.
-            BTreeMap::new(memory, max_key_size, max_value_size)
+            BTreeMap::new(memory)
         } else {
             // The memory already contains a BTreeMap. Load it.
             BTreeMap::load(memory)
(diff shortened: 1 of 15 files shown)


ielashi:
@frederikrothenberger @hpeebles Adding you to this PR as users of `StableBTreeMap` to get your opinions.

src/storable.rs
```diff
@@ -37,6 +43,12 @@ impl Storable for () {
     }
 }
 
+impl BoundedStorable for () {
```
hpeebles, line 46: Maybe use a macro for all of these numeric type implementations.
Also is there a reason i8, i16, etc aren't covered?

hpeebles:
I think this is a good simplification which will cover lots of use cases.
But the `BoundedStorable` trait is only actually needed when the map is being created via the new ctor.
It may be useful for some use cases to still expose another ctor which doesn't require `BoundedStorable` where you can specify your own limits.

ielashi:
> But the `BoundedStorable` trait is only actually needed when the map is being created via the new ctor. It may be useful for some use cases to still expose another ctor which doesn't require `BoundedStorable` where you can specify your own limits.

Yeah, I see your point. The idea was that you could wrap the type you want to use, just like I did [in the example](https://github.com/dfinity/stable-structures/pull/17/files#diff-0ff8ed2711e7694d032ad0a64ca532cfa5e50361ed5b4c19b16cddb4e0f1500eR35).

We can potentially introduce something like `init_with_sizes` that allows you to specify the sizes explicitly to preserve the previous usage pattern.

src/btreemap.rs
```diff
@@ -928,15 +943,22 @@ mod test {
...
 
+    // Make `Vec<u8>` bounded so that it can be used as a key/value in the btree.
+    impl BoundedStorable for Vec<u8> {
+        fn max_size() -> u32 {
+            10
+        }
+    }
+
     #[test]
     fn init_preserves_data() {
```
frederikrothenberger, line 954: Maybe I missed it, but shouldn't there be a (new) test somewhere that checks that the `BTreeMap` will complain if you try to load from memory but the bounds of the types provided are too large?

frederikrothenberger (approved):
I really like this change 👍 
Left a few comments.

src/btreemap.rs
```diff
@@ -125,6 +128,18 @@ impl<M: Memory + Clone, K: Storable, V: Storable> BTreeMap<M, K, V> {
         let header: BTreeHeader = read_struct(Address::from(0), &memory);
         assert_eq!(&header.magic, MAGIC, "Bad magic.");
         assert_eq!(header.version, LAYOUT_VERSION, "Unsupported version.");
+        let expected_key_size = header.max_key_size;
+        assert_eq!(
+            expected_key_size,
+            K::max_size(),
```
dsarlis, line 134:
```suggestion
            max_key_size,
```

src/btreemap.rs
```diff
@@ -125,6 +128,18 @@ impl<M: Memory + Clone, K: Storable, V: Storable> BTreeMap<M, K, V> {
...
+            "Expected key size doesn't match actual key size"
+        );
+        let expected_value_size = header.max_value_size;
+        assert_eq!(
+            expected_value_size,
+            V::max_size(),
```
dsarlis, line 140:
```suggestion
            max_value_size,
```

dsarlis (approved):

src/storable.rs
```diff
@@ -15,6 +15,12 @@ pub trait Storable {
     fn from_bytes(bytes: Vec<u8>) -> Self;
 }
 
+/// A trait indicating that a `Storable` element is bounded in size.
+pub trait BoundedStorable: Storable {
```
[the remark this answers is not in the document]
ielashi: That and tuples are on my list :)

src/storable.rs
```diff
@@ -15,6 +15,12 @@ pub trait Storable {
...
+    /// The maximum size, in bytes, of the type when serialized.
+    fn max_size() -> u32;
```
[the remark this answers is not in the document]
ielashi: I agree, but currently the `max_size` that `StableBTreeMap` stores is u32, so we'll have to use the same type here unfortunately.

src/storable.rs
ielashi: Totally, but one step at a time :)

src/btreemap.rs
```diff
@@ -125,6 +128,18 @@ impl<M: Memory + Clone, K: Storable, V: Storable> BTreeMap<M, K, V> {
         let header: BTreeHeader = read_struct(Address::from(0), &memory);
         assert_eq!(&header.magic, MAGIC, "Bad magic.");
         assert_eq!(header.version, LAYOUT_VERSION, "Unsupported version.");
+        let expected_key_size = header.max_key_size;
```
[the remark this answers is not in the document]
ielashi: Done.

src/btreemap.rs
ielashi: I was admittedly being lazy. Thanks for calling me out.

ielashi (commented):
Thanks for the great feedback. I addressed all the comments, and added a `(init|load|new)_with_sizes` to preserve the former API for those who happen to prefer it.

PR #18, opened 2022-11-23
Add basic stable vec implementation

fooris:


fooris:
This is a basic port of the rust standard library implementation (std::vec)

src/vec.rs
```diff
@@ -0,0 +1,296 @@
...
+            self.reserve(1);
+        }
+
+        let p = self.index_to_addr(index);
+        let src = p;
+        let dst = p + Bytes::from(self.max_value_size as u64);
+        copy::<T, M>(&self.memory, src, dst, (len - index) as u64);
+        write_struct(element, p, &self.memory);
+
+        self.len += 1;
```
roman-kashitsyn, line 142: Don't we need to flush the header to memory at this point?

src/vec.rs
```diff
@@ -0,0 +1,296 @@
...
+
+        // space for the new element
+        if len == self.capacity() {
+            self.reserve(1);
+        }
+
+        let p = self.index_to_addr(index);
+        let src = p;
+        let dst = p + Bytes::from(self.max_value_size as u64);
+        copy::<T, M>(&self.memory, src, dst, (len - index) as u64);
```
roman-kashitsyn, line 139: I'm not sure I understand what we're doing here.

src/vec.rs
```diff
@@ -0,0 +1,296 @@
...
+        write_struct(element, p, &self.memory);
```
roman-kashitsyn, line 140: I think this design misses the point of stable structures. We need to _serialize_ values of T to disk using Storable::to_bytes/Storable::from_bytes methods. Otherwise, recompiling your canister or adding a new optional field to a structure can invalidate all your data.

src/vec.rs
```diff
@@ -0,0 +1,296 @@
...
+    use std::rc::Rc;
+
+    fn make_memory() -> Rc<RefCell<Vec<u8>>> {
+        Rc::new(RefCell::new(Vec::new()))
+    }
+
+    #[test]
+    pub fn test_insert() {
+        let mem = make_memory();
+        let mut v = StableVec::new(mem);
```
roman-kashitsyn, line 246: We should also test that the vector works if we drop it and re-load from `mem`.

src/vec.rs
fooris: Ah, it seems i misunderstood the write_struct method.
I thought it would serialize the struct and write it to memory.
On closer inspection, this was false.

Maybe some kind of comment on write_struct that makes its use clear would be helpful.

src/vec.rs
fooris: yes, we do. I wanted to ask whether the data structure itself should flush, or if the user should do it.
This comment answers that question, thx!

src/vec.rs
fooris: Well, the default behavior of insert is that it will add an entry at position 'index'.
so if v = [0,1,2,3]
and we do insert(1, 10)
we get v = [0,10,1,2,3].

For this we need to move [1,2,3] 1 over to the right. Which is what we do here.

I realize that this code is not very self explanatory, so i'll add a comment.

src/storable.rs
```diff
@@ -21,6 +21,10 @@ pub trait BoundedStorable: Storable {
     fn max_size() -> u32;
 }
 
+pub trait EncodableStorable: BoundedStorable {
+    const FIXED_LEN: bool;
```
ielashi, line 25: I'd put this as part of the BoundedStorable trait. No need for a separate trait imo.

fooris: Done.

src/vec.rs
```diff
@@ -0,0 +1,364 @@
...
+        // Move all entries starting at index one to the right.
+        let src = self.index_to_entry_addr(index);
+        let dst = self.index_to_entry_addr(index + 1);
+        copy(
+            &self.memory,
+            src,
+            dst,
+            (len - index) as u64,
+            Self::ENTRY_SIZE,
+        );
```
ielashi, line 169: The amount of memory that needs to be shifted here is unbounded. If I have a vector containing millions of entries, this call may need to move gigabytes of state. We have a limit on how much computation we can do per message (measured in a unit called "instructions"), and this kind of shifting with a large enough state would cause whatever is currently running to panic.

src/vec.rs
```diff
@@ -0,0 +1,364 @@
...
+        let ret = self.read_value(index);
+
+        // Shift everything down to fill in that spot.
+        copy::<M>(
+            &self.memory,
+            self.index_to_entry_addr(index + 1),
+            self.index_to_entry_addr(index),
+            (len - index - 1) as u64,
+            Self::ENTRY_SIZE,
+        );
```
ielashi, line 190: Same concern about shifting. Here, one potential solution would be to mark an element as deleted, and whenever the user provides an `index`, we modify the `index` to go to the appropriate location based on which elements have been deleted.

ielashi (commented):
Thanks for the contribution @fooris. My primary concern with this data structure is the shifting we are doing in `insert` and `remove`, which is not something we can do due to the instruction limits we have for canisters.

src/vec.rs
fooris: see https://github.com/dfinity/stable-structures/pull/18/files/970473e8601434c19f146d7cbf341a52750eeee1#r1033744554

src/vec.rs
fooris: Actually, that is exactly what we need for the binary heap.

fooris: But we still need need instructions to access (read/overwrite) elements within the vector. So i guess it would still not quite be a stack.

src/vec.rs
```diff
@@ -0,0 +1,364 @@
...
+        }
+    }
+
+    /// # Safety
+    ///
+    /// Calling this method with an out-of-bounds index is *[undefined behavior]*
+    /// even if the resulting reference is not used.
+    pub unsafe fn get_unchecked(&self, index: usize) -> T {
+        self.read_value(index)
+    }
```
ielashi, line 221: What is the use-case for this endpoint?

src/vec.rs
```diff
@@ -0,0 +1,364 @@
...
+        }
+    }
+
+    pub fn set(&mut self, index: usize, value: &T) {
+        assert!(index < self.len());
+        self.write_value(index, value);
+    }
+
+    pub fn reserve(&mut self, additional: usize) {
+        let page_size = 64 * 1024;
```
ielashi, line 237: nit: we can use the `WASM_PAGE_SIZE` constant defined in this crate.

src/vec.rs
```diff
@@ -0,0 +1,364 @@
...
+
+        write_struct(&header, Address::from(0), &self.memory);
+    }
+
+    pub fn clear(&mut self) {
+        self.len = 0;
+        self.save();
+    }
+
+    #[inline]
```
ielashi, line 120: I'm not sure I understand why we're inlining some of these methods. This is a private method, so I'd expect the compiler to know whether or not to inline it. Is this based on some benchmarking?

src/vec.rs
```diff
@@ -0,0 +1,364 @@
...
+        assert!(index < self.len());
+        self.write_value(index, value);
+    }
+
+    pub fn reserve(&mut self, additional: usize) {
+        let page_size = 64 * 1024;
+        let round_up = |x| ((x + page_size - 1) & !(page_size - 1));
+
+        // We need to reserve space for the u32 that indicates the size.
+        let add_bytes = (additional * Self::ENTRY_SIZE) as u64;
```
ielashi, line 241: I wonder if we're better off using `leb128` encoding for the size to cut down on the number of bytes needed.

src/vec.rs
ielashi: Fair enough. A growable array with `push` and `pop` does seem generic enough, and I can use-cases for this even without the `insert` and `remove`. So, yes, I'd suggest removing these APIs.

src/vec.rs
fooris: If I know that the access in bounds. For example in the binary heap, I do the exact same check.

src/vec.rs
fooris: Done.

saikatdas0790:
Any movement on this? This would be an awesome addition to this library?

roman-kashitsyn:
Contributing to the original PR was too hard, so I re-implemented stable vector from scratch here: https://github.com/dfinity/stable-structures/pull/23

PR #22, opened 2022-12-23
Fix #15 - make MemoryManager::init_with_buckets public

### LINKED ISSUE #15: Big preallocation slows tests
In cargo tests, we want to provide independent memory, memory manager and data structures for each test. But for first allocation `MemoryManager` allocate a bucket with 1024 pages of memory (about 67Mb). That operation is slow and requires a lot of memory, which will never be used.
I think a library client need some control over first allocation size. For example, client may replace `const BUCKET_SIZE_IN_PAGES`, with custom value in `MemoryManager` constructor.

ufoscout:
Fixes #15

### DIFF
--- a/src/memory_manager.rs
+++ b/src/memory_manager.rs
@@ -139,7 +139,8 @@ impl<M: Memory> MemoryManager<M> {
         Self::init_with_buckets(memory, BUCKET_SIZE_IN_PAGES as u16)
     }
 
-    fn init_with_buckets(memory: M, bucket_size_in_pages: u16) -> Self {
+    /// Initializes a `MemoryManager` with the given memory and bucket size in pages.
+    pub fn init_with_buckets(memory: M, bucket_size_in_pages: u16) -> Self {
         Self {
             inner: Rc::new(RefCell::new(MemoryManagerInner::init(
                 memory,


roman-kashitsyn (approved):

ielashi:
IIUC the reason we want to make this public is that unit tests are currently slow, since every memory used by the `MemoryManager` will allocate at least 64MiB.

The 64MiB bucket size was chosen somewhat arbitrarily. I'm wondering, if instead of exposing the bucket size in an API and moving that complexity on the user if it'd be simpler if we instead shrink the default bucket size from 64MiB to, say, 8MiB. I think that would suffice for production and not be a big burden in unit tests. WDYT? cc @roman-kashitsyn

ielashi:
I looked into this more closely and added benchmarks to understand the impact of the bucket size of the `MemoryManager` on performance (see #29). Based on the data, it seems that a smaller bucket size has a negligible impact on performance, so it'd make sense based on the data to reduce the current default bucket size from 1024 pages to 64 or 128 pages. A smaller default bucket size would make issue #15 obsolete, I believe.

Having said that, I don't think there's any reason not to give developers the ability to specify a custom bucket size, so this PR still LGTM.

ielashi (approved):

PR #23, opened 2022-12-29
feat: add stable vec implementation

roman-kashitsyn:
This change adds a new stable structure, Vec, modeling a growable array in stable memory.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -96,7 +96,8 @@ impl Display for GrowFailed {
         write!(
             f,
             "Failed to grow memory: current size={}, delta={}",
-            self.current_size, self.delta
+            self.current_size,
+            self.delta
         )
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,6 +20,7 @@ pub mod storable;
 mod tests;
 mod types;
 pub mod vec;
+pub use vec::{Vec as StableVec, Vec};
 pub mod vec_mem;
 pub mod writer;
 pub use btreemap::{BTreeMap, BTreeMap as StableBTreeMap};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,6 +19,7 @@ pub mod storable;
 #[cfg(test)]
 mod tests;
 mod types;
+pub mod vec;
 pub mod vec_mem;
 pub mod writer;
 pub use btreemap::{BTreeMap, BTreeMap as StableBTreeMap};
@@ -95,8 +96,7 @@ impl Display for GrowFailed {
         write!(
             f,
             "Failed to grow memory: current size={}, delta={}",
-            self.current_size,
-            self.delta
+            self.current_size, self.delta
         )
     }
 }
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -19,6 +19,7 @@ pub mod storable;
 #[cfg(test)]
 mod tests;
 mod types;
+pub mod vec;
```
ielashi, line 22: nit: maybe expose `Vec` directly here as well?

src/vec.rs
```diff
@@ -0,0 +1,324 @@
...
+        })
+    }
+
+    /// Initializes a vector in the specified memory.
+    ///
+    /// Complexity: O(1)
+    ///
+    /// PRECONDITION: the memory is either empty or contains a valid
+    /// stable vector.
+    pub fn init(memory: M) -> Result<Self, InitError> {
```
ielashi, line 115: Curious, do you think the `InitError` returned here and the `GrowFailed` elsewhere are recoverable at runtime? If not, then why return the error as opposed to using asserts?

ielashi (approved):
Looks great, thanks!

src/vec.rs
roman-kashitsyn: > why return the error as opposed to using asserts?

I see two reasons why typed errors are better in `new` and `init`:
1. Explicit errors make testing better.
2. The user can attach a custom error message with an `expect`. This ability is essential if you have multiple data structures of the same kind. If one of them blows up, you want to know which one it was. Data structure constructors don't have enough context.

Furthermore, I think `Storable` would benefit from error handling in `from_bytes` as well so that the data structure code could report the context where the error happened, something like "failed to decode entry at index 5: ...".

src/lib.rs
roman-kashitsyn: Good idea, done!

src/vec.rs
ielashi: I agree that creating stable structures is rare and it's ok to add extra verbosity there. Two follow-ups:

1. This PR also adds errors to the main API. The `push` method returns a `GrowFailed`, and I don't expect the use of the `push` API to be rare. The inconvenience there can add up, especially that, even if the user knows which specific call to `push` lead to growing the memory to fail, there's still nothing they can do about it.

2. Why is `set` an exception? The error in `set` is arguably more likely to occur than the error in `push`, and it's something the user can/should protect against.

I'd suggest making assertions in the `push` API that growing the memory succeeds so that it's more convenient and more consistent with `set`, but that's not a strong opinion, so feel free to move forward with whatever option appeals to you.

roman-kashitsyn: > Why is set an exception? The error in set is arguably more likely to occur than the error in push, and it's something the user can/should protect against.

`set` can panic only due to a programming error when the user messes up the indices (just like regular vector assignment). `push` returns an error under regular operation when the memory fills up.
The `push` error is also recoverable: for example, and archive canister can signal to the main canister that it's full so it's time to create another archive. I can't imagine a case where `set` error would be recoverable.

PR #24, opened 2022-12-29
feat: add benchmark for BTreeMap

ielashi:
Adds a benchmark for `BTreeMap` insertions and includes those in CI. More benchmarks and better CI tooling will be added in upcoming commits.


ielashi:
After some more investigation, this benchmark is not representative of the performance on the IC. I'll use a canister instead for benchmarking to give us more accurate/relevant data.

PR #25, opened 2022-12-30
chore: update owners

ielashi:


ielashi:
@dsarlis do you know if we can add individuals to owners? I have a vague memory hearing that we should only add teams, but I'm not entirely sure. Do you recall?

ielashi:
I asked around but didn't get a concrete answer. I'll merge this to not lose the momentum of our work and if this ends up violating some policy then we can adjust accordingly later.

PR #26, opened 2022-12-30
optimize: reduce stable memory accesses in BTreeMap by 10x.

ielashi:
Loading and saving nodes in `BTreeMap` required many accesses to the underlying memory. With 4 separate accesses to the underlying memory for every element in the node, a full node can require over 40 calls to `memory.read` or `memory.write`.

The code has been updated to first read/write to a `VectorMemory` (i.e. on the heap), and then read/write the node in one single call. Below is the output of `cargo bench`. The performance improvement speaks for itself.

```
BTreeMap::insert        time:   [22.391 ms 22.603 ms 22.891 ms]
                        change: [-88.996% -88.849% -88.689%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  6 (6.00%) high mild
  4 (4.00%) high severe
```

### DIFF
--- a/src/btreemap/node.rs
+++ b/src/btreemap/node.rs
@@ -1,8 +1,9 @@
 use crate::{
     read_struct, read_u32, read_u64,
     types::{Address, Bytes},
-    write, write_struct, write_u32, Memory,
+    write, write_struct, write_u32, Memory, VectorMemory,
 };
+use std::{cell::RefCell, rc::Rc};
 
 /// The minimum degree to use in the btree.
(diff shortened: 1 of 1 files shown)


ielashi:
The benchmark that this optimization was based on was not representative of real performance on the IC. When I ran this patch on a real canister, performance actually declined significantly.

I'll first add benchmarks that use a real canister to get accurate readings on what can be optimized.

PR #27, opened 2022-12-31
feat: add btreemap benchmarks

ielashi:
Adds a `benchmarks` canister that is used for benchmarking btreemap insertions/removals. Future commits will include:

* More benchmarks
* More measurements (e.g. number of dirty pages)
* CI integration

Benchmarks can be run using `./benchmarks/run.sh`.


roman-kashitsyn (approved):
LGTM!

Because of this change, you'll have to adjust the benchmark code a bit: https://github.com/dfinity/stable-structures/pull/28.

ielashi:
For reference, here are the benchmark numbers:

```
(831_679_957 : nat64)
(1_559_244_733 : nat64)
```

PR #28, opened 2023-01-01
refactor(btreemap): change the type argument order

roman-kashitsyn:
This change changes the order of the type arguments in BTreeMap to be more consistent with other data structures and with the standard library (Allocators are the last type arguments in Rust).

### DIFF
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -2222,11 +2222,4 @@ mod test {
         // Equal key size
         let _btree: BTreeMap<Vec<u8>, Vec<u8>, _> = BTreeMap::init_with_sizes(mem, 4, 3);
     }
-
-    #[test]
-    fn supports_unbounded_storable_types() {
-        let mem = make_memory();
-        // `String` doesn't implement `BoundedStorable` but can still be used in the `BTreeMap`.
-        let _btree: BTreeMap<String, String, _> = BTreeMap::init_with_sizes(mem.clone(), 4, 3);
-    }
 }
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -2222,4 +2222,11 @@ mod test {
         // Equal key size
         let _btree: BTreeMap<Vec<u8>, Vec<u8>, _> = BTreeMap::init_with_sizes(mem, 4, 3);
     }
+
+    #[test]
+    fn supports_unbounded_storable_types() {
+        let mem = make_memory();
+        // `String` doesn't implement `BoundedStorable` but can still be used in the `BTreeMap`.
+        let _btree: BTreeMap<String, String, _> = BTreeMap::init_with_sizes(mem.clone(), 4, 3);
+    }
 }
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -5,7 +5,7 @@ mod node;
 use crate::{
     read_struct,
     types::{Address, Bytes, NULL},
-    write_struct, BoundedStorable, Memory, Storable,
+    write_struct, BoundedStorable, Memory,
 };
 use allocator::Allocator;
 pub use iter::Iter;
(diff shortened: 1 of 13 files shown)


src/btreemap.rs
```diff
@@ -400,9 +396,13 @@ impl<M: Memory + Clone, K: Storable, V: Storable> BTreeMap<M, K, V> {
         self.length
     }
 
-    /// Returns a reference to the memory used by the map.
-    pub fn get_memory(&self) -> M {
-        self.memory.clone()
+    /// Returns the underlying memory.
+    pub fn forget(self) -> M {
+        self.allocator.forget()
+    }
```
ielashi, line 402: I think the name `forget` here can be misleading, since no data is actually forgotten. Consider this code:

```
let btree: BTreeMap<u64, u64. _> = BTree::init(mem);
btree.insert(1,1).unwrap();
let mem = btree.forget();
let btree: BTreeMap<u64, u64. _> = BTree::init(mem);
assert_eq!(btree.len(), 0); // panics
```

Given the name `forget`, I'd assume that this deletes the btree from memory, and it leads to the, IMO, the unintuitive result of a newly initialized btree not being being empty.

Another example:

```
let btree = BTree::init_with_sizes(10, 10, mem);
let mem = btree.forget();
let btree = BTree::init_with_sizes(20, 20, mem); // will panic
```

I'd also, judging by the name `forget`, assume that my initial btree doesn't exist anymore and I can initialize a new one in its place, but that's not what will happen. The underlying btree is still intact, and trying to re-initialize a tree with larger keys/values panics.

src/btreemap.rs
```diff
@@ -20,7 +20,7 @@ const MAGIC: &[u8; 3] = b"BTR";
 ///
 /// The implementation is based on the algorithm outlined in "Introduction to Algorithms"
 /// by Cormen et al.
-pub struct BTreeMap<M: Memory, K: Storable, V: Storable> {
+pub struct BTreeMap<K: BoundedStorable, V: BoundedStorable, M: Memory> {
```
ielashi, line 23: The use of `Storable` here, and not `BoundedStorable`, is deliberate (ditto everywhere else), since users can still use the `BTreeMap` with unbounded `Storable` types. It's a flaw in our test suite that this change didn't break any tests.

src/btreemap.rs
roman-kashitsyn: Here, `forget` means "forget that you have a map in get me back the raw memory". If it was supposed to delete the contents, I'd call it "clear()" or "erase()`.
We use this name in other data structures as well. Do you think `into_memory()` would be a better name?

ielashi: I think `into_memory` is much clearer, thanks!

ielashi:
LGTM after swapping the `BoundedStorable` trait with `Storable` per my comment. I added a small test case to this PR to ensure we keep it that way.

ielashi (approved):

src/btreemap.rs
```diff
@@ -2191,35 +2191,42 @@ mod test {
...
         // Equal key size
-        let _btree: BTreeMap<_, Vec<u8>, Vec<u8>> = BTreeMap::init_with_sizes(mem, 4, 3);
+        let _btree: BTreeMap<Vec<u8>, Vec<u8>, _> = BTreeMap::init_with_sizes(mem, 4, 3);
+    }
+
+    #[test]
+    fn supports_unbounded_storable_types() {
+        let mem = make_memory();
+        // `String` doesn't implement `BoundedStorable` but can still be used in the `BTreeMap`.
+        let _btree: BTreeMap<String, String, _> = BTreeMap::init_with_sizes(mem.clone(), 4, 3);
```
roman-kashitsyn, line 2230: We should address the issue of using strings as keys by adding a bounded string type, `Text<CAP>`. This will make interfaces across data structures more consistent and will communicate the intent more clearly.

src/btreemap.rs
ielashi: That's a fair point. I am currently exploring removing the bound altogether, but it's not clear whether or not this exploration will be fruitful, so, until then, it would make sense to make using `BoundedStorable` required, as it does capture the intent better than specifying the sizes in the constructor. We should remove the `init_with_sizes` API in this case though, but that I'm happy to do in a follow-up PR.

PR #29, opened 2023-01-04
feat: add benchmarks for `MemoryManager`.

ielashi:
To investigate #15 , this PR adds benchmarks to measure the overhead of using a `MemoryManager` with various bucket sizes compared to using the underlying memory directly. The results:

```
+ dfx canister call benchmarks memory_manager_baseline --query
(1_176_576_470 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(64:nat16)' --query
(1_182_016_996 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(128:nat16)' --query
(1_181_977_011 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(512:nat16)' --query
(1_181_945_444 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(1024:nat16)' --query
(1_181_939_464 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(2048:nat16)' --query
(1_181_936_474 : nat64)
```

1. The overhead of using the `MemoryManager`, compared to using the underlying memory directly, is around 10%.
2. Surprisingly, the bucket size of the `MemoryManager` seems to have little effect on performance. This would justify reducing the current default bucket size of 1024 to a smaller size as it would result in a lower memory footprint.


roman-kashitsyn (approved):
LGTM

PR #33, opened 2023-01-06
feat(btreemap): implement proper range scan

roman-kashitsyn:
This change replaces the `range` method based on prefixes with a `range` method based on key range, similarly to the `BTreeMap::range` from the standard library.

To achieve this, we:

  1. Require map keys to implement Ord.
  2. Store node keys decoded.
  3. Store the key bounds in the map iterator.

I also separated keys and values into separate arrays, because it makes the code a bit cleaner and usually a bit more efficient.

### DIFF
--- a/examples/src/vecs_and_strings/src/lib.rs
+++ b/examples/src/vecs_and_strings/src/lib.rs
@@ -13,7 +13,7 @@ type Memory = VirtualMemory<DefaultMemoryImpl>;
 const MAX_USER_NAME_SIZE: u32 = 10;
 const MAX_USER_DATA_SIZE: u32 = 100;
 
-#[derive(PartialOrd, Ord, Clone)]
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone)]
 struct UserName(String);
 
 impl Storable for UserName {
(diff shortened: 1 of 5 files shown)


ielashi:
API changes LGTM, thanks for the simplification and the additional clarity.

Looking at our benchmarks, I'm noticing a performance regression of ~10%. I think this is mostly related to the changes in `node.rs`. For example, now we deserialize every single key in the node, and that introduces overhead. Maybe this is something that we can avoid?

Before:
```
(831_639_971 : nat64)
(1_559_704_733 : nat64)
```

After:
```
(927_805_429 : nat64)
(1_716_531_037 : nat64)
```

roman-kashitsyn:
>  Looking at our benchmarks, I'm noticing a performance regression of ~10%. I think this is mostly related to the changes in node.rs. 

Good catch!

> For example, now we deserialize every single key in the node, and that introduces overhead. Maybe this is something that we can avoid?

I believe decoding keys might be actually faster than storing raw vectors. Especially if the keys are fixed-size, such as arrays or integers. 

In [this commit 
](https://github.com/dfinity/stable-structures/pull/33/commits/94a9dded5eab32a2801fe864b56ebb61bdd74038), I use optimization opportunities opened up https://github.com/dfinity/stable-structures/pull/35. I read keys into a pre-allocated buffer and decode from borrowed bytes. The results are quite impressive for a 5-line change:

```
(609_084_581 : nat64)
(1_083_249_173 : nat64)
```

ielashi:
That is indeed impressive. LGTM!

ielashi (approved):

PR #35, opened 2023-01-08
refactor(storable): eliminate unnecessary buffer clones

roman-kashitsyn:
This change modifies the `Storable::from_bytes` method to accept `Cow<[u8]>` instead of a byte vector. Often, a storable type does not need to own the original byte array (the exceptions are only `Vec<u8>` and `String`).

There are a few places (such as iterators) where we can eliminate unnecessary byte array clones with this optimization.

The new signature is also more symmetrical:
`T::from_bytes(t.to_bytes())` becomes valid code.


ielashi (approved):
LGTM, thanks!

PR #37, opened 2023-01-12
chore(btreemap): remove errors from 'insert' API

ielashi:
`BTreeMap::insert` returned an `InsertError` if the key/value inserted was too large. Now that keys/values are required to implement `BoundedStorable`, the only scenario where an `InsertError` can be triggered is if the implementation of `Storable` and/or `BoundedStorable` for the key/value has a bug.

This commit changes the API to remove the `InsertError`, replacing it with a precondition that assumes that the implementations of `Storable`/`BoundedStorable` are correct. This has the advantage that the API is now cleaner, and panicking whenever we detect a bug in the implementation is safer for the developer.


roman-kashitsyn (approved):
LGTM, nice improvement!

PR #38, opened 2023-01-13
Enforce the ordering of the fields

dfinity-berestovskyy:
Note, packed does not provide any guarantee about field ordering, and dereferencing un unaligned pointers is an undefined behavior.


roman-kashitsyn (approved):
LGTM, thanks!

PR #39, opened 2023-01-17
Remove BTreeHeader representation attribute

dfinity-berestovskyy:
Dereferencing an unaligned pointer is undefined behavior
and it is possible to safely create unaligned pointers to packed fields.

### DIFF
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -74,7 +74,7 @@ where
     _phantom: PhantomData<(K, V)>,
 }
 
-// The total header size must be <= ALLOCATOR_OFFSET
+// The total (packed) header size must fit before the allocator.
 struct BTreeHeaderV1 {
     magic: [u8; 3],
     version: u8,
@@ -174,8 +174,10 @@ where
 
     /// Reads the header from the specified memory.
     fn read_header(memory: &M) -> BTreeHeaderV1 {
-        // The header must fit before the allocator
-        debug_assert!(core::mem::size_of::<BTreeHeaderV1>() < ALLOCATOR_OFFSET as usize);
+        // The total (packed) header size must fit before the allocator.
+        // Due to the padding, at some point the size of BTreeHeaderV1
+        // might get bigger, so this assert should be fixed.
+        debug_assert!(core::mem::size_of::<BTreeHeaderV1>() <= ALLOCATOR_OFFSET as usize);
         // Read the header
         let mut buf = [0; core::mem::size_of::<BTreeHeaderV1>()];
         memory.read(0, &mut buf);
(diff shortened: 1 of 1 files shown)


src/btreemap/allocator.rs
```diff
@@ -45,7 +45,7 @@ pub struct Allocator<M: Memory> {
     memory: M,
 }
 
-#[repr(packed)]
+#[repr(C, packed)]
```
ielashi, line 48: Maybe we can keep this change (and other similar changes) out of this PR as it isn't part of updating the `BTreeHeader` implementation? This change will also be obsolete once we migrate these to use the same pattern as `BTreeHeader`, correct?

src/btreemap.rs
```diff
@@ -60,6 +58,9 @@ struct BTreeHeader {
     _buffer: [u8; 24],
 }
 
+/// Current header size with additional header space
+const HEADER_SIZE: usize = 52;
```
ielashi, line 62: Now that we have this constant, I think it'd useful to document the header layout just like we do [here](https://github.com/dfinity/stable-structures/blob/main/src/vec.rs#L3). WDYT?

src/btreemap/allocator.rs
dfinity-berestovskyy: This PR is just built on top of https://github.com/dfinity/stable-structures/pull/38 The idea was to merge `#[repr(C, packed)]` ASAP, and then agree on the unaligned pointers in this MR...

src/btreemap.rs
```diff
@@ -1,11 +1,35 @@
...
+//! ----------------------------------------
+//! Layout version         ↕ 1 byte
+//! ----------------------------------------
+//! Max key size           ↕ 4 bytes
+//! ----------------------------------------
+//! Max value size         ↕ 4 bytes
+//! ----------------------------------------
+//! Root node address      ↕ 8 bytes
+//! ----------------------------------------
+//! Number of nodes        ↕ 8 bytes
```
ielashi, line 18:
```suggestion
//! Length (number of elements)        ↕ 8 bytes
```

src/btreemap.rs
```diff
@@ -934,17 +965,30 @@ where
...
+        // Serialize the header
+        let mut buf = [0; core::mem::size_of::<BTreeHeaderV1>()];
+        buf[0..3].copy_from_slice(&header.magic);
+        buf[3] = header.version;
+        buf[4..8].copy_from_slice(&header.max_key_size.to_le_bytes());
+        buf[8..12].copy_from_slice(&header.max_value_size.to_le_bytes());
+        buf[12..20].copy_from_slice(&header.root_addr.get().to_le_bytes());
+        buf[20..28].copy_from_slice(&header.length.to_le_bytes());
+        // Write the header
+        crate::safe_write(memory, 0, &buf)
```
ielashi, line 991: Since we're simply unwrapping the value returned here, we could use the `write` method that panics when growing fails.

```suggestion
        crate::write(memory, 0, &buf)
```

ielashi (approved):
LGTM! I am also noticing a little performance improvement with this work.

Before:

```
(609_514_567 : nat64)
(1_083_249_173 : nat64)
```

After:

```
(585_114_282 : nat64)
(1_039_926_728 : nat64)
```

PR #41, opened 2023-01-19
chore: update library version

ielashi:
This is in preparation for the upcoming release.


roman-kashitsyn (approved):
LGTM, but it seems that we need to fix the examples...

PR #43, opened 2023-01-20
chore: refactor MemoryManager layout implementation

dfinity-berestovskyy:
Dereferencing an unaligned pointer is undefined behavior
and it is possible to safely create unaligned pointers to packed fields.

The performance is slightly improved thanks to the `grow()` optimization. Before:
```
+ dfx canister call benchmarks memory_manager_baseline --query
(1_176_576_470 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(64:nat16)' --query
(1_182_016_996 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(128:nat16)' --query
(1_181_977_011 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(512:nat16)' --query
(1_181_945_444 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(1024:nat16)' --query
(1_181_939_464 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(2048:nat16)' --query
(1_181_936_474 : nat64)
```
After:
```
+ dfx canister call benchmarks memory_manager_baseline --query
(1_176_576_470 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(64:nat16)' --query
(1_177_230_726 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(128:nat16)' --query
(1_177_190_441 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(512:nat16)' --query
(1_177_158_649 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(1024:nat16)' --query
(1_177_152_619 : nat64)
+ dfx canister call benchmarks memory_manager_interleaved '(2048:nat16)' --query
(1_177_149_604 : nat64)
```


benchmarks/run.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 set -Eexuo pipefail
 
+export DFX_WARNING=-version_check
```
ielashi, line 4: I'm not sure we want to suppress this warning. Different versions of dfx can report different instruction counts, so it's important to use the latest version for all benchmarks IMHO.

Would you mind also adding what the benchmarks numbers are before/after this change? Note that #40 needs to be merged before you're able to do so.

src/btreemap.rs
```diff
@@ -16,9 +16,9 @@
 //! Root node address           ↕ 8 bytes
 //! ----------------------------------------
 //! Length (number of elements) ↕ 8 bytes
-//! ----------------------------------------
+//! ---------------------------------------- <- Address 28 (PACKED_HEADER_SIZE)
```
ielashi, line 19: IIUC the changes in this file are unrelated to the MemoryManager changes right? Maybe worth updating the description accordingly or splitting this out into its own PR?

ielashi:
Looks great, thanks @dfinity-berestovskyy!

benchmarks/run.sh
dfinity-berestovskyy: It's the opposite, there is a newer beta `dfx`, but I'm fine to drop it. #40 was re-implemented in this MR in fact, so completely dropping [this commit](https://github.com/dfinity/stable-structures/pull/43/commits/31f77ce8f514447c8f419ff15d3fac2e3730a90d)... 🤣

src/btreemap.rs
dfinity-berestovskyy: Split into #47

PR #46, opened 2023-01-23
chore: remove unnecessary representation attributes

dfinity-berestovskyy:
There is no need for representation of single field structure.

### DIFF
--- a/src/btreemap/allocator.rs
+++ b/src/btreemap/allocator.rs
@@ -166,6 +166,7 @@ impl<M: Memory> Allocator<M> {
         chunk.save(chunk_addr, &self.memory);
 
         // Update the head of the free list.
+        #[allow(unaligned_references)]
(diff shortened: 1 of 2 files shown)


ielashi:
I am also unclear here what the problem is. Is it that it's wrong to have a representation of a single field structure, or is it that you find the code more clear without it?

dfinity-berestovskyy:
> I am also unclear here what the problem is. Is it that it's wrong to have a representation of a single field structure, or is it that you find the code more clear without it?

I thought it would be nice to get rid of all the representation attributes, so this is just a cleanup.

ielashi (approved):
@dfinity-berestovskyy and I discussed this offline. There will be a follow-up PR that migrated `ChunkHeader` to use direct reads/writes from/to memory rather than a `repr` struct.

PR #47, opened 2023-01-23
chore: minor B-Tree fixes

dfinity-berestovskyy:
* Added `create_btreemap_dump_file()` to update the binary file for the future layout changes.
* Use `PACKED_HEADER_SIZE` instead of `size_of::<BTreeHeaderV1>()` for now, as the `size_of` of unpacked structure is a bit an overshot.
* Other minor changes


ielashi:
The changes LGTM, but I'm a bit unclear as to what these changes fix exactly. Do you mind elaborating in the PR description?

dfinity-berestovskyy:
> The changes LGTM, but I'm a bit unclear as to what these changes fix exactly. Do you mind elaborating in the PR description?

Added a bit more  details. In general, I'm not sure about this solution is the best one, but it definitely more correct than the previous `size_of`...

IMO  the approach we took previously with packed structures was in a way a nice solution. Maybe, adding just `u8` arrays inside the packed struct is the solution. WDYT?

ielashi (approved):

PR #48, opened 2023-01-23
fix(log, vec): use u64 for indices and lengths

roman-kashitsyn:
On the Wasm32 architecture, usize is 32 bit, making it impossible for the log to report correct size in bytes on large data sets.

This commit changes the types of indices and lengths in Vec and Log so that they will behave correctly if these structures grow beyond four billion items.

### DIFF
--- a/src/log.rs
+++ b/src/log.rs
@@ -256,18 +256,18 @@ impl<T: Storable, INDEX: Memory, DATA: Memory> Log<T, INDEX, DATA> {
     }
 
     /// Returns the number of index memory bytes in use.
-    pub fn index_size_bytes(&self) -> usize {
+    pub fn index_size_bytes(&self) -> u64 {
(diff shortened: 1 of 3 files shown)


ielashi (approved):
LGTM. We'll need to release a new minor version `0.5.0`, to roll this out, right?

roman-kashitsyn:
> LGTM. We'll need to release a new minor version `0.5.0`, to roll this out, right?

Do you think we should do a similar change for the BTreeMap?

ielashi:
> > LGTM. We'll need to release a new minor version `0.5.0`, to roll this out, right?
> 
> Do you think we should do a similar change for the BTreeMap?

We already use a `u64` in `BTreeMap::len()`. Do you see anything else that needs to be updated?

roman-kashitsyn:
> We already use a u64 in BTreeMap::len(). Do you see anything else that needs to be updated?

No, I think we're good. Thanks!

PR #50, opened 2023-01-31
feat: add benchmarks for BTreeMap

ielashi:
# Problem
The benchmarks we have aren't comprehensive enough to get an accurate picture of the performance of `BTreeMap`. For example, in the `0.4.0` release, the benchmarking numbers showed us an improvement of ~40% in the `insert` and `remove` operations, but when I profiled the Bitcoin canister with that upgrade I saw a _regression_ of about 40%.

# Solution
Add more thorough benchmarks for `insert`, `remove`, and `get` to get a more accurate picture of the performance impact of our changes. These will then be used to address the performance regressions we've seen with `0.4.0+`

### DIFF
--- a/benchmarks/run.sh
+++ b/benchmarks/run.sh
@@ -12,14 +12,40 @@ trap "dfx stop" ERR EXIT
 dfx start --background --clean
 dfx deploy --no-wallet benchmarks
 
-# BTreeMap benchmarks
-dfx canister call benchmarks btreemap_insert --query
-dfx canister call benchmarks btreemap_remove --query
-
 # MemoryManager benchmarks
 dfx canister call benchmarks memory_manager_baseline --query
-dfx canister call benchmarks memory_manager_interleaved "(64:nat16)" --query
-dfx canister call benchmarks memory_manager_interleaved "(128:nat16)" --query
-dfx canister call benchmarks memory_manager_interleaved "(512:nat16)" --query
-dfx canister call benchmarks memory_manager_interleaved "(1024:nat16)" --query
-dfx canister call benchmarks memory_manager_interleaved "(2048:nat16)" --query
+dfx canister call benchmarks memory_manager_overhead --query
+
+# BTreeMap benchmarks
+dfx canister call benchmarks btreemap_insert_blob_4_1024 --query
+dfx canister call benchmarks btreemap_insert_blob_8_1024 --query
+dfx canister call benchmarks btreemap_insert_blob_16_1024 --query
+dfx canister call benchmarks btreemap_insert_blob_32_1024 --query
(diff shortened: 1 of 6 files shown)


benchmarks/run.sh
```diff
@@ -12,14 +12,40 @@ trap "dfx stop" ERR EXIT
...
+
+dfx canister call benchmarks btreemap_get_blob_4_1024 --query
+dfx canister call benchmarks btreemap_get_blob_8_1024 --query
+dfx canister call benchmarks btreemap_get_blob_16_1024 --query
+dfx canister call benchmarks btreemap_get_blob_32_1024 --query
+dfx canister call benchmarks btreemap_get_blob_64_1024 --query
+dfx canister call benchmarks btreemap_get_blob_128_1024 --query
+# These tests go over the instruction limit, so we can't run them currently.
+#dfx canister call benchmarks btreemap_get_blob_256_1024
+#dfx canister call benchmarks btreemap_get_blob_512_1024
```
ielashi, line 40: In a future PR, I plan to use a version of `drun` to allow us to run these benchmarks without a cycles limit.

dsarlis: They go over even with the increased DTS limits?

benchmarks/src/btreemap.rs
```diff
@@ -1,47 +1,252 @@
...
+use ic_cdk_macros::{query, update};
+use ic_stable_structures::{storable::Blob, BTreeMap, BoundedStorable, DefaultMemoryImpl};
 use tiny_rng::{Rand, Rng};
 
-/// Benchmarks inserting keys into a BTreeMap.
-#[ic_cdk_macros::query]
-pub fn btreemap_insert() -> u64 {
-    let mut btree: BTreeMap<u64, (), _> = BTreeMap::init(DefaultMemoryImpl::default());
+#[query]
+pub fn btreemap_insert_blob_4_1024() -> u64 {
```
dsarlis, line 7: nit: We can probably reduce all this boilerplate with a simple macro that defines the functions for each combination.

dsarlis (approved):
> For example, in the 0.4.0 release, the benchmarking numbers showed us an improvement of ~40% in the insert and remove operations, but when I profiled the Bitcoin canister with that upgrade I saw a regression of about 40%.

Ouch...

benchmarks/run.sh
ielashi: The replica that ships with 0.12.1 has a limit of 10B, and that's enough for us at the moment.

benchmarks/src/btreemap.rs
ielashi: True. This code will evolve quite a bit in the next commits though, so I want to first wait until the code is more stable before working on removing the boilerplate.

PR #53, opened 2023-02-03
feat(btreemap): add methods to get first/last entry

roman-kashitsyn:
This change adds new methods to the StableBTreeMap interface:

  1. clear() removes everything from the map.
  2. first_key_value() returns the entry with the smallest key.
  3. last_key_value() returns the entry with the largest key.

### DIFF
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -39,6 +39,9 @@ use std::borrow::Cow;
 use std::marker::PhantomData;
 use std::ops::{Bound, RangeBounds};
 
+#[cfg(test)]
+mod proptests;
+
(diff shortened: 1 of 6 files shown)


src/btreemap.rs
```diff
@@ -39,6 +39,9 @@ use std::borrow::Cow;
 use std::marker::PhantomData;
 use std::ops::{Bound, RangeBounds};
 
+#[cfg(test)]
+mod proptests;
```
ielashi, line 43: I assume this file is missing from the commit?

roman-kashitsyn: Ouch, fixed.

ielashi (approved):

PR #60, opened 2023-03-06
feat(btreemap): add iter_upper_bound api

roman-kashitsyn:
This change introduces a new btree map method, `iter_upper_bound`. It returns an iterator pointing to the first element that is less than the given key.

We need this method for the time series API because we need to find the closest keys to the left and to the right of the given key. We can do the latter with the `range` API, but the former requires a new method.

We took inspiration from experimental API in BTreeMap: https://doc.rust-lang.org/nightly/std/collections/struct.BTreeMap.html#method.upper_bound

### DIFF
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -942,17 +942,8 @@ where
         loop {
             match node.keys.binary_search(bound) {
                 Ok(idx) | Err(idx) => {
-                    let child = match node.node_type {
-                        NodeType::Internal => {
-                            // Note that loading a child node cannot fail since
-                            // len(children) = len(entries) + 1
-                            Some(self.load_node(node.children[idx]))
-                        }
-                        NodeType::Leaf => None,
-                    };
-
-                    match child {
-                        None => {
+                    match node.node_type {
+                        NodeType::Leaf => {
                             if idx > 0 {
                                 cursors.push(Cursor::Node {
                                     node,
@@ -970,14 +961,16 @@ where
                                 return Iter::null(self);
                             }
                         }
-                        Some(child) => {
+                        NodeType::Internal => {
+                            let child = self.load_node(node.children[idx]);
+
                             if idx < node.keys.len() {
                                 cursors.push(Cursor::Node {
                                     node,
                                     next: Index::Entry(idx),
                                 });
                             }
-                            // Iterate over the child node.
+
                             node = child;
                         }
                     }
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -930,6 +930,62 @@ where
         }
     }
 
+    /// Returns an interator pointing to the first element below the specified key.
+    pub fn iter_upper_bound(&self, bound: &K) -> Iter<K, V, M> {
+        if self.root_addr == NULL {
+            // Map is empty.
+            return Iter::null(self);
+        }
+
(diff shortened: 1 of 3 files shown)


src/btreemap.rs
```diff
@@ -2388,6 +2444,18 @@ mod test {
         }
     }
 
+    #[test]
+    fn test_iter_upper_bound() {
+        let mut stable_map = super::BTreeMap::new(make_memory());
+        for k in 0..1000u64 {
+            stable_map.insert(k, ());
+            println!("Getting a upper bound for {}", k + 1);
```
dsarlis, line 2452: Debug leftover?

src/btreemap.rs
```diff
@@ -930,6 +930,62 @@ where
...
+            // Map is empty.
+            return Iter::null(self);
+        }
+
+        let mut cursors = vec![];
+        let mut node = self.load_node(self.root_addr);
+        loop {
+            match node.keys.binary_search(bound) {
+                Ok(idx) | Err(idx) => {
+                    let child = match node.node_type {
```
dsarlis, line 945: Do you really need to define `child`? We can probably just inline the code you have below in one match right away.

dsarlis (approved):

src/btreemap.rs
roman-kashitsyn: Done!

src/btreemap.rs
roman-kashitsyn: These prints are helpful for the case when the test fails. Turned those into better assert messages.

src/btreemap.rs
```diff
@@ -930,6 +930,56 @@ where
...
+                                });
+                                // Leaf node. Return an iterator with the found cursors.
+                                return Iter::new_in_range(
+                                    self,
+                                    (Bound::Unbounded, Bound::Unbounded),
+                                    cursors,
+                                );
+                            } else {
+                                // The upper bound is less than or equal to the first key in the map.
+                                // We return an empty iterator.
```
roman-kashitsyn, line 961: This logic is too simplistic. The problem with upper bounds is that we might have to backtrack.
If we see an internal node with keys [1, 10] and look for upper bound 5, we might descend into the slot between 1 and 10 and find keys [6, 8, 9] there, meaning that we should have taken 1.

That never happens with `range`, so we need a slightly different approach for the upper bound.

src/btreemap.rs
```diff
@@ -930,6 +930,56 @@ where
         }
     }
 
+    /// Returns an interator pointing to the first element below the given bound.
```
dsarlis, line 933:
```suggestion
    /// Returns an iterator pointing to the first element below the given bound.
```

src/btreemap.rs
dsarlis: So you mean there's a bug really here? We shouldn't return null but rather keep looping through so that we catch 1 in your example?

src/btreemap.rs
roman-kashitsyn: Good catch, fixed!

src/btreemap.rs
roman-kashitsyn: Yes, there was a bug. I wrote a more thorough test (checking upper bounds for _all_ keys, not just the last key) and found the bug. It's fixed now!

PR #63, opened 2023-03-10
feat: add a constant for the max number of pages

roman-kashitsyn:
This change introduces a new constant, MAX_PAGES, indicating the maximum number of pages a canister can address. This constant is useful for constructing a restricted memory that can span up to the last page of the base memory.

```rust
RestrictedMemory::new(memory, 16..MAX_PAGES)
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -43,6 +43,9 @@ pub type DefaultMemoryImpl = VectorMemory;
 
 const WASM_PAGE_SIZE: u64 = 65536;
 
+/// The maximum number of stable memory pages a canister can address.
+pub const MAX_PAGES: u64 = u64::MAX / WASM_PAGE_SIZE;
+
 pub trait Memory {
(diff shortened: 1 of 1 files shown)


dsarlis (approved):

src/lib.rs
```diff
@@ -181,7 +184,7 @@ pub struct RestrictedMemory<M: Memory> {
 
 impl<M: Memory> RestrictedMemory<M> {
     pub fn new(memory: M, page_range: core::ops::Range<u64>) -> Self {
-        assert!(page_range.end < u64::MAX / WASM_PAGE_SIZE);
+        assert!(page_range.end <= MAX_PAGES);
```
dsarlis, line 187: Is it intentional you include the equality?

maksymar (approved):

src/lib.rs
roman-kashitsyn: Yes, because the right bound in Range is excluded.

PR #67, opened 2023-03-26
feat: add principal and bool types

0kenx:


src/storable.rs
```diff
@@ -297,6 +298,39 @@ impl<const N: usize> BoundedStorable for [u8; N] {
...
+        Cow::Owned(self.as_slice().to_vec())
+    }
+
+    fn from_bytes(bytes: Cow<[u8]>) -> Self {
+        Self::from_slice(bytes.as_ref())
+    }
+}
+
+impl BoundedStorable for Principal {
+    const MAX_SIZE: u32 = 30;
```
dsarlis, line 312: Isn't the max size of a principal 29 bytes as per the [interface spec](https://internetcomputer.org/docs/current/references/ic-interface-spec#principal)?

Cargo.toml
```diff
@@ -18,6 +18,9 @@ maplit = "1.0.2"
 proptest = "1"
 tempfile = "3.3.0"
 
+[dependencies]
+candid = "0.8.4"
```
ielashi, line 22: It has been an intentional design decision to not introduce dependencies to this library unless absolutely necessary. In this case, I'm not sure if forcing all users to now depend on candid is the right call. I think we have three options here:

1. Not include this dependency, in which case, users that want to store principals will need to wrap it by defining a `struct WrapperPrincipal(Principal)` in their own crate.
2. We include this dependency, making all users now depend on candid.
3. Add this as an optional feature, so that it can imported by users that need it.

It seems like approach #3 here would be most appropriate. WDYT? cc @dsarlis @roman-kashitsyn

src/storable.rs
ielashi, line 312: The max size of a principal is 29 bytes.

ielashi:
@0kenx thanks a lot for your contribution! I left a couple of comments.

src/storable.rs
```diff
@@ -297,6 +298,39 @@ impl<const N: usize> BoundedStorable for [u8; N] {
...
+
+impl Storable for bool {
+    fn to_bytes(&self) -> Cow<[u8]> {
+        Cow::Borrowed(if *self { &[1] } else { &[0] })
+    }
+
+    fn from_bytes(bytes: Cow<[u8]>) -> Self {
+        match bytes.get(0) {
+            Some(0) | None => false,
+            _  => true,
```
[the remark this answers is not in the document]
ielashi: +1. Test cases for these scenarios would be great as well.

Cargo.toml
0kenx: Normally I would prefer the approach #3 as well. However I would imagine most people would need to use principals in stable memory, and almost everyone developing on the IC is already dependent on candid in any case.

src/storable.rs
0kenx: it's 29 bytes, plus 1 byte for the length.

Cargo.toml
dsarlis: I tend to lean towards option 3 because not everyone might want to use candid to encode what they store in stable memory (there are already cases where people opt for something different because of better performance or more flexible representation for stable data). Having it behind a feature still makes it available relatively easily for the people that want to use candid.

ielashi: I think @0kenx brought up fair arguments though. It's virtually impossible to not depend on candid when building a Rust canister, unless you build a Rust canister that doesn't expose a candid interface, and it's safe to say that these are very, very rare.

The candid dependency isn't related to how people choose to encode/decode their stable data - it's to implement `Storable` for the `Principal` type. I'm starting to see some appeal behind option 1 - most people will need this, given how primitive the `Principal` type is to the IC. In the future, it'd be nice if we move the `Principal` type into a separate crate so that we don't have to depend on the entire candid library.

dsarlis: My opinion was not strong, indeed the argument about having a candid interface is a good one. I'm fine with including the dependency too.

0kenx: FYI I'm using a custom newtype in my application now. I'm just saying that if we agree that `principal` and ICRC-1 `(principal, subaccount)` are first-class citizens in the IC, then they should have first-class library support.

src/storable.rs
0kenx: https://github.com/dfinity/candid/blob/0ae3c22c6df17f7e09fb49c5ce5fe77653884e10/rust/candid/src/types/principal.rs#L85

0kenx: But you're right. `as_slice` doesn't include the length.

Cargo.toml
ielashi: @0kenx I think that's reasonable. Let's go ahead with that (i.e. stick with option 1 as is done in this PR).

dsarlis:
@0kenx any updates here? Is there anything that needs further clarification? Or is it just simply a matter of finding the cycles to get back to this?

0kenx:
Do we agree to add `candid` as a dependency here? Sorry it wasn't immediately clear from our discussion above.

ielashi:
> Do we agree to add `candid` as a dependency here? Sorry it wasn't immediately clear from our discussion above.

@0kenx  Yes, we did.

0kenx:
Added `candid` as a feature so that dependency is no longer mandatory.

dsarlis (approved):

Cargo.toml
```diff
@@ -18,6 +18,9 @@ maplit = "1.0.2"
 proptest = "1"
 tempfile = "3.3.0"
 
+[dependencies]
+candid = { version = "0.8.4", optional = true}
```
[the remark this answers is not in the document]
0kenx: It is implemented as a feature, so the dependency is optional.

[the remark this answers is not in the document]
roman-kashitsyn: That's also not ideal because it complicates the development and the infrastructure.
* We must bump the candid versions when new versions appear (0.9 is coming soon, and it's a breaking change).
* We create a dependency between candid and stable structure versions that can make migrating to newer versions of `stable-structures` unnecessarily burdensome.
* Developers of `stable-structues` need to compile and test with features enabled; we must adjust the CI to test the entire feature matrix.

I don't think it's a fair bargain to win a single type conversion.

PR #77, opened 2023-05-12
fix: use drun for running the benchmarks.

ielashi:
# Problem
Benchmarks are run within a canister to obtain the instruction count. However, benchmarks can run for up to 5B instructions in case of queries, and 50B in case of updates. We do have benchmarks that we haven't been able to run because they surpass these limits.

# Solution
Use `drun` instead of `dfx` to run the benchmarks. `drun` can be configured to have a practically infinite instruction count. As a result, a few benchmarks are now enabled by this commit, and all the benchmarks are now query calls to avoid side effects.

Closes EXC-1405

### DIFF
--- a/benches/run-benchmark.sh
+++ b/benches/run-benchmark.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+#
+# Runs a benchmark using drun. The script assumes drun is available on the caller's path.
+set -euo pipefail
+
+BENCH_NAME=$1
+FILE=mktemp
+
+if ! type "drun" > /dev/null; then
+  echo "drun is not installed. Please add drun to your path from commit d35535c96184be039aaa31f68b48bbe45909494e."
+  exit 1
+fi
+
+cat > $FILE << EOF
+create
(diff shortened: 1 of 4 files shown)


benches/run-benchmark.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+#
+# Runs a benchmark using drun. The script assumes drun is available on the caller's path.
+set -euo pipefail
+
+BENCH_NAME=$1
+FILE=mktemp
+
+if ! type "drun" > /dev/null; then
+  echo "drun is not installed. Please add drun to your path from commit d35535c96184be039aaa31f68b48bbe45909494e."
```
roman-kashitsyn, line 10: It would be great if we could run these benchmarks on each push request and push the outcome as a comment.
Or at least have a CI action that runs the benchmarks.

roman-kashitsyn (approved):

benches/run-benchmark.sh
ielashi: Absolutely. We have a [separate ticket for that](https://dfinity.atlassian.net/browse/EXC-1406), which I'll take care as part of the on-going stable structures feature.

PR #84, opened 2023-06-01
[POC] BTreeMap with unbounded keys and values.

ielashi:
This commit is a proof of concept for a new version of `BTreeMap` that doesn't require keys or values to be bounded in size. This POC is a simplified and inefficient implementation, and I'd like to get feedback on the general direction of this approach. Specifically:

1. Adding `BOUND` to `Storable`
  Because Rust doesn't support generic specializations, we cannot have special logic in the `BTreeMap` based on whether or not the keys/values implement `BoundedStorable`. To get around this, I propose merging `BoundedStorable` into `Storable` by introducing the optional `BOUND` field. Structures can then support both bounded and unbounded types, but with the downside being that bound checks would now happen at run-time.
  
2. Introduce variable-length nodes.
  Nodes in the `BTreeMap` have been fixed in size, where the maximum possible size is allocated for each key and value. In the V2 proposed in this POC, nodes can be variable in size, growing and shrinking to accommodate its contents. A node still contains the same bounds on the number of entries as in V1, and that allows us to reuse the same BTreeMap logic, which remains practically untouched.
  
    A node is initially given a `page_size`, and when it needs to outgrow it, it does so by allocating additional overflow pages of the same size. Having everything grow by a fixed page allows us to reuse the constant-size chunk `Allocator` with no modifications.

The inefficiency of the current implementation is due to it loading and writing all the entries on every load/save. A production-grade implementation would make these loads lazy.

**Advantages of this approach**
- Minimal to no change to the BTreeMap logic.
- No change to the BTreeMap's allocator.
- No change to the BTreeMap API.
- V1 versions of the map can be upgraded to V2 seamlessly

**Disadvantages of this approach**
- Requires specifying a `page_size`. A `page_size` that's too large will waste memory, and a `page_size` that's too small will lead to a lot of overflow pages. If the developer specifies bounds for keys and/or values, then these can be used as heuristics to set an appropriate `page_size`.
- There is some complexity with supporting lazy saving of entries, where only the dirtied entries are written to memory. Because the sizes of entries are unbounded, we'll need something like a first-fit allocator to rewrite dirtied entries within the page efficiently.


ielashi:
Having done a bit more research, I'm not sure that removing the `BoundedStorable` trait in favor of adding `BOUND` to the `Storable` trait is the best approach. Not only do bound checks now need to happen at runtime, but it's also impossible to compose bounded types, like we do with [tuples](https://github.com/dfinity/stable-structures/blob/main/src/storable.rs#L315). The alternative would be to introduce a new type, say `UnboundedBTreeMap`. It's unfortunate, but it seems to be the cleaner, more type-safe approach.

lastmjs:
Would the developer need to or be able to set the `page_size`? Ideally this complexity would not be exposed to the developer.

What options do we have for minimizing the impact of the `page_size`, for example could the `page_size` start out at a reasonable default and adapt itself over time?

src/btreemap/node.rs
```diff
@@ -1,18 +1,24 @@
...
+mod v2;
 
 // The minimum degree to use in the btree.
 // This constant is taken from Rust's std implementation of BTreeMap.
 const B: usize = 6;
 // The maximum number of entries per node.
 const CAPACITY: usize = 2 * B - 1;
-const LAYOUT_VERSION: u8 = 1;
+const LAYOUT_VERSION_1: u8 = 1;
+const LAYOUT_VERSION_2: u8 = 2;
```
dragoljub-duric, line 21: I guess with the new approach `UnboundedBTreeMap`, these versions will disappear. Am I right?

src/btreemap/node/v1.rs
```diff
@@ -0,0 +1,152 @@
+use super::*;
```
dragoljub-duric, line 1: Is the whole of this file, just old methods for BoundedBTreeMap moved to one place?

src/btreemap/node/v2.rs
```diff
@@ -0,0 +1,305 @@
...
+            address,
+            keys,
+            encoded_values: RefCell::new(encoded_values),
+            children,
+            node_type,
+            version: Version::V2 { page_size },
+            overflow: if original_overflow_address == crate::types::NULL {
+                None
+            } else {
+                Some(original_overflow_address)
```
dragoljub-duric, line 94: When loading the `Node`, why we do not load the overflow data as well? Is that avoided to decrease refactoring when lazy loading is done or there is another reason?

ielashi:
> Would the developer need to or be able to set the `page_size`? Ideally this complexity would not be exposed to the developer.

No, this would be set internally by the BTreeMap, but we could potentially provide an option to the developer to set it, should they be a power user.

> What options do we have for minimizing the impact of the `page_size`, for example could the `page_size` start out at a reasonable default and adapt itself over time?

As things stand, where we are using a constant-size chunk allocator, once a `page_size` is set, it cannot be changed. I'm exploring an alternative design to this POC where we instead use a dynamic allocator. I'll post more on that soon.

src/btreemap/node.rs
ielashi: Yes, correct.

src/btreemap/node/v1.rs
ielashi: I wouldn't consider them old methods. I'd think of them more as shared functionality between both btrees.

src/btreemap/node/v2.rs
ielashi: Yes, exactly, we want to do as little work as possible here.

lastmjs:
I'm very interested in the dynamic allocator, sounds great.

ielashi:
Closing as all these changes have been incorporated and merged into main in other PRs.

PR #86, opened 2023-06-01
chore: update dependencies in examples

peterpeterparker:
# Motivation

Few (24) days ago a new version of [ciborium](https://crates.io/crates/ciborium) was released.
Therefore we can use the official dependency in the quickstart example instead of referencing a particular commit.

In addition, this PR bumps the candid and ic-* dependencies of the examples as provided by @ielashi.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.20"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


ielashi:
Looks like the examples are not passing. You need to regenerate the Cargo.lock file there.

peterpeterparker:
@ielashi you mean in `./examples/src`? I get an error

```
❯ cargo update
    Updating crates.io index
error: failed to select a version for the requirement `ic-cdk = "^0.4"`
candidate versions found which didn't match: 0.8.1, 0.8.0, 0.7.4, ...
```

While I'm there, should I bump all dependencies of the examples?

ielashi (approved):

PR #91, opened 2023-06-14
feat: stable min heap data structure

roman-kashitsyn:
This change introduces the `MinHeap` stable structure analogous to `std::collections::BinaryHeap`.
Contrary to `BinaryHeap`, `MinHeap` prioritizes smaller items, which I found much more useful in practice than prioritizing larger items.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -97,7 +97,8 @@ impl Display for GrowFailed {
         write!(
             f,
             "Failed to grow memory: current size={}, delta={}",
-            self.current_size, self.delta
+            self.current_size,
+            self.delta
         )
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,7 @@ mod ic0_memory; // Memory API for canisters.
 pub mod log;
 pub use log::{Log as StableLog, Log};
 pub mod memory_manager;
+pub mod min_heap;
 pub mod reader;
 pub mod storable;
 #[cfg(test)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,7 +16,7 @@ pub mod storable;
 mod tests;
 mod types;
 pub mod vec;
-pub use min_heap::MinHeap as StableMinHeap;
+pub use min_heap::{MinHeap, MinHeap as StableMinHeap};
 pub use vec::{Vec as StableVec, Vec};
 pub mod vec_mem;
 pub mod writer;
@@ -97,8 +97,7 @@ impl Display for GrowFailed {
         write!(
             f,
             "Failed to grow memory: current size={}, delta={}",
-            self.current_size,
-            self.delta
+            self.current_size, self.delta
         )
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,6 +16,7 @@ pub mod storable;
 mod tests;
 mod types;
 pub mod vec;
+pub use min_heap::MinHeap as StableMinHeap;
 pub use vec::{Vec as StableVec, Vec};
 pub mod vec_mem;
 pub mod writer;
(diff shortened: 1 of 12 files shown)


src/min_heap/tests.rs
```diff
@@ -0,0 +1,130 @@
...
+#[derive(Debug, PartialEq, Clone)]
+enum Operation<T> {
+    Push(T),
+    Pop,
+}
+
+fn arb_op<T: Clone + Debug>(s: impl Strategy<Value = T>) -> impl Strategy<Value = Operation<T>> {
+    prop_oneof![
+        3 => s.prop_map(Operation::Push),
+        1 => Just(Operation::Pop),
```
ielashi, line 19: What is the purpose of the keys `3` and `1` here? Removing them doesn't seem to have any effect.

```suggestion
        s.prop_map(Operation::Push),
        Just(Operation::Pop),
```

src/lib.rs
```diff
@@ -9,12 +9,14 @@ mod ic0_memory; // Memory API for canisters.
...
 pub use log::{Log as StableLog, Log};
 pub mod memory_manager;
+pub mod min_heap;
 pub mod reader;
 pub mod storable;
 #[cfg(test)]
 mod tests;
 mod types;
 pub mod vec;
+pub use min_heap::MinHeap as StableMinHeap;
```
ielashi, line 19: nit: should we also export the name without the stable prefix, as we do with other types?

```suggestion
pub mod vec;
pub use min_heap::{MinHeap as StableMinHeap, MinHeap};
```

src/min_heap.rs
```diff
@@ -0,0 +1,210 @@
...
+mod tests;
+
+const MAGIC: [u8; 3] = *b"SMH"; // Short for "stable min heap".
+
+/// An implementation of a binary heap.
+///
+/// Contrary to [std::collections::BinaryHeap], this heap is a min-heap (smallest items come first).
+/// Motivation: max heaps are helpful for sorting, but most daily programming tasks require min
+/// heaps.
+pub struct MinHeap<T: BoundedStorable + PartialOrd, M: Memory>(BaseVec<T, M>);
```
ielashi, line 16: nit: shall we add this to the list of available data structures in the [readme](https://github.com/dfinity/stable-structures#available-data-structures)?

ielashi (approved):
Nice and simple, thanks!

ielashi:
Maybe also adding an example to the examples folder would be nice?

src/min_heap.rs
```diff
@@ -0,0 +1,210 @@
...
+
+#[cfg(test)]
+mod tests;
+
+const MAGIC: [u8; 3] = *b"SMH"; // Short for "stable min heap".
+
+/// An implementation of a binary heap.
+///
+/// Contrary to [std::collections::BinaryHeap], this heap is a min-heap (smallest items come first).
+/// Motivation: max heaps are helpful for sorting, but most daily programming tasks require min
```
AlexandraZapuc, line 14: I would remove the motivation comment, and simply specify that this is an implementation of a `Min Binary Heap`. I would leave the explanation only in the description.

AlexandraZapuc (approved):

src/min_heap/tests.rs
roman-kashitsyn: These numbers are weights affecting the probabilities. I want pushes to be 3x more often so that the heap has a chance to grow relatively large.

src/lib.rs
roman-kashitsyn: Done.

src/min_heap.rs
roman-kashitsyn: Done.

src/min_heap.rs
roman-kashitsyn: Done. I've also added a changelog file.

roman-kashitsyn:
> Maybe also adding an example to the examples folder would be nice?

I've added an example, PTAL.

Issue #69, opened 2023-04-11
Remove the BoundedStorable requirement


lastmjs:
Currently all values that are to be stored in a stable structure (as far as I know) require the `BoundedStorable`  trait to be implemented. This requires the developer to reason about possible maximum and fixed sizes of all of their data types, recursively, and is a significant hurdle to the developer experience of stable structures.

At Demergent Labs we have exposed TypeScript and Python versions of the stable structures library to developers, and we plan to use stable memory, and hope to use stable structures, under-the-hood to implement our sudodb/sudograph databases. If we have to implement the `BoundedStorable` trait for the databases, we will be exposing significant complexity to our end users.

Not to mention the efficiency concerns. It would be great to have this requirement lifted in the future, which would be a boon to Azle, Kybra, and Sudograph.

ielashi:
Hey @lastmjs! We were just having discussions about the roadmap for stable structures, and this is definitely one of the higher priority changes we'll be working on.

lastmjs:
Any update on this?

ielashi:
@lastmjs Yes! I was off for the past two weeks. We do have a POC now in https://github.com/dfinity/stable-structures/pull/84, and I'll work on getting a production-ready version of it merged in over the coming days.

lastmjs:
Thanks for the update, very excited

lastmjs:
Just updating this thread, I'm testing this locally in Azle and so far all is well. Once this is officially released I'd say we can close this issue.

lastmjs:
Amazing!!!

PR #94, opened 2023-06-23
chore: merge `BoundedStorable` into `Storable`

ielashi:
## Problem
We'd like to expand the functionality of `BTreeMap` to support unbounded types. To add this support in a backward-compatible way, the `BTreeMap` would need to support both bounded types (i.e. types that implement `BoundedStorable`) and unbounded types (i.e. types that don't implement `BoundedStorable`).

Rust doesn't support generic specializations, so we cannot have the same `BTreeMap` support both use-cases simultaneously. It can either support only `BoundedStorable` types, or all types implementing `Storable`, but without the knowledge of whether a `BoundedStorable` implementation is available for those types.

## Solution
Merge `BoundedStorable` and `Storable` into the same trait. The one downside of this approach is that checking whether or not a bound exists now happens at run-time.

This impacts the developer experience. Prior to the PR, if you try to store a `String` inside a `StableVec`, the compiler will complain that `String` doesn’t implement `BoundedStorable`. But, with this solution, it’ll happily compile, and only when you run your code will you see a panic that `String` is not bounded.

This PR relates to #84 and #69.


src/storable.rs
```diff
@@ -15,21 +15,30 @@ pub trait Storable {
 
     /// Converts bytes into an element.
     fn from_bytes(bytes: Cow<[u8]>) -> Self;
+
+    /// The size bounds the type.
```
dsarlis, line 19:
```suggestion
    /// The size bounds of the type.
```

dragoljub-duric:
I am more in favor of the second approach with the separate `UnboundedBTreeMap`:

1. The first approach with the check in the run-time in my opinion does not go in line with what most libraries in the Rust offer. And I think that we can shoot ourselves in the leg with that approach.
2. When the user uses `UnboundedBTreeMap` for the type that implements `BoundedStorable` it is his error, and he will pay the price for it. I see it more or less as the difference between `map` and `unordered_map` (in programming languages) because you can do everything with the `map` as you can do with the `unordered_map` with some additional things, but if you do not need those additional things you may waste the resources using `map`.

ielashi:
> 1. The first approach with the check in the run-time in my opinion does not go in line with what most libraries in the Rust offer. I am afraid that we can shoot ourselves in the leg with that approach.

What is an example of us "shooting ourselves in the leg"?

PR #95, opened 2023-06-27
feat: Support memory freeing

dragoljub-duric:
Problem:
Stable structures do not support memory freeing.

Solution:
To free memory, we’d need to add the following API to the memory manager:

`fn free(memory: MemoryId) -> ();
`
Post Conditions:
- The size of memory is zero.
- The buckets used by memory are unallocated.
- Currently, the memory manager allocates new buckets by growing its underlying memory. Instead, we should ensure that allocations of new buckets by the memory manager use all the unallocated buckets of memory first before growing the underlying memory.

To do:
Maybe in the future remove the external dependency BitVec and use bit operations instead.


src/memory_manager.rs
```diff
@@ -248,37 +313,82 @@ impl<M: Memory> MemoryManagerInner<M> {
             memory_sizes_in_pages: [0; MAX_NUM_MEMORIES as usize],
             memory_buckets: BTreeMap::new(),
             bucket_size_in_pages,
+            unallocated_buckets: {
+                let mut list = LinkedList::new();
+                for i in 0..MAX_NUM_BUCKETS {
+                    list.push_back(BucketId(i as u16));
+                }
+                list
+            },
```
ielashi, line 322: It seems a bit wasteful that we create a list of 32k elements here. How about, instead of tracking all of the `unallocated_buckets`, we only track the `freed_buckets`? When a memory needs to grow, we first use the buckets in the `freed_buckets`, and if that's empty, we allocate new buckets, as we do in V1. WDYT?

src/memory_manager.rs
```diff
@@ -419,6 +561,43 @@ impl<M: Memory> MemoryManagerInner<M> {
...
+                self.allocated_buckets -= 1;
+            }
+        }
+        // Update the header.
+        self.save_header();
+
+        // Write in stable store that no bucket belongs to the memory with the provided `id`.
+        write(
+            &self.memory,
+            bucket_allocations_address(BucketId(0)).get(),
```
ielashi, line 580: I think the use `bucket_allocations_address(BucketId(0))` here and elsewhere is misleading. The function `bucket_allocations_address` is specific to V1, where it returns the address where we store the allocation of the given bucket. That's not how they're being used in this code.

I think what your code truly cares about is the offset where the bucket indexes are stored. So instead of using `bucket_allocations_address(BucketId(0))`, we instead can define a constant, say `BUCKET_INDEXES_OFFSET` with that value.

src/memory_manager.rs
```diff
@@ -419,6 +561,43 @@ impl<M: Memory> MemoryManagerInner<M> {
...
     }
+
+    fn free(&mut self, id: MemoryId) {
+        self.memory_sizes_in_pages[id.0 as usize] = 0;
+        let buckets = self.memory_buckets.remove(&id);
+        if let Some(vec_buckets) = buckets {
+            for bucket in vec_buckets {
+                self.unallocated_buckets.push_front(bucket);
+                self.allocated_buckets -= 1;
+            }
```
ielashi, line 572: Let's say we have two memories:

```
Memory 1: Buckets[1, 2, 3],
Memory 2: Buckets[4, 5, 6],
```

Then I do the following:

```
free(Memory 1)
free(Memory 2)

get(Memory 3)
grow(Memory 3, 3)
```

IIUC the code above will first use buckets 4, 5, 6 before using 1, 2, 3. This doesn't matter too much right now, but it might matter if in the future we add defragmentation. It's more desirable to use the buckets in order (i.e. 1, 2, 3, then 4, 5, 6. Does that make sense?

src/memory_manager.rs
```diff
@@ -853,4 +1032,104 @@ mod test {
         memory_1.read(0, &mut buf);
         assert_eq!(buf, vec![2; 1000]);
     }
+
```
ielashi, line 1035: Have you considered the following scenarios:

1. upgrading the memory manager from V1 to V2
2. freeing an already free memory
3. freeing a memory that was never used
4. freeing a memory, then growing it

This is a big change, and we need to have solid test coverage.

src/memory_manager.rs
```diff
@@ -853,4 +1032,104 @@ mod test {
...
+        // Check that data of MemoryId 0 and 2 is correctly reinitialized.
+        memory_0.read(0, &mut bytes);
+        assert_eq!(bytes, vec![1, 2, 3]);
+
+        memory_2.read(0, &mut bytes);
+        assert_eq!(bytes, vec![7, 8, 9]);
+
+        // Check that date of MemoryId 1 is freed.
+        assert_eq!(memory_1.size(), 0);
+        memory_1.read(0, &mut bytes);
```
ielashi, line 1133: Can this test not simply be? It's not clear to me what the additional logic adds to the test.

```
       let mem = make_memory();
       let mem_mgr = MemoryManager::init(mem.clone());
       let memory_0 = mem_mgr.get(MemoryId(0));
       mem_mgr.free(MemoryId(0));
       memory_0.read(0, &mut bytes);
```

ielashi:
Thanks @dragoljub-duric. I left some comments. Additionally, the PR description I think can be improved, where we concisely write the problem and the solution. It's currently more of a checklist on how to implement this PR.

src/memory_manager.rs
```diff
@@ -287,13 +439,103 @@ impl<M: Memory> MemoryManagerInner<M> {
...
+            }
+        } else {
+            let first_unused_bucket = BucketId((last_occupied_bucket + 1) as u16);
+
+            let mut freed_buckets = LinkedList::new();
+            for i in unallocated_buckets.iter() {
+                if *i < first_unused_bucket.0 {
+                    freed_buckets.push_back(BucketId(*i));
+                }
+            }
```
ielashi, line 498: Aren't we here still storing all the unallocated buckets up to `MAX_NUM_BUCKETS`? Per my comment yesterday, that's not necessary and it's more efficient to store only the buckets that were previously freed (i.e. buckets with ids <= last_occupied_bucket)

src/memory_manager.rs
```diff
@@ -419,6 +665,143 @@ impl<M: Memory> MemoryManagerInner<M> {
...
     }
+
+    fn free(&mut self, id: MemoryId) {
+        self.memory_sizes_in_pages[id.0 as usize] = 0;
+        let buckets = self.memory_buckets.remove(&id);
+        if let Some(vec_buckets) = buckets {
+            for bucket in vec_buckets {
+                self.freed_buckets.push_back(bucket);
+                self.allocated_buckets -= 1;
+            }
```
ielashi, line 676: As discussed before, we should reuse freed buckets  in sorted order (i.e. freed bucket `n` should be used before freed bucket `n - 1`). Changing `push_front` to `push_back` doesn't change that. Maybe using a BTreeSet would be more appropriate here.

src/memory_manager.rs
ielashi: Yes, that would be preferable, because in the future we could increase the number of buckets, and in that case it's not desirable to allocate a collection that's `MAX_NUM_BUCKETS` in size.

src/memory_manager.rs
```diff
@@ -526,6 +933,10 @@ fn bucket_allocations_address(id: BucketId) -> Address {
     Address::from(0) + Header::size() + Bytes::from(id.0)
 }
 
+fn bucket_indexes_offset() -> Address {
+    Address::from(0) + Header::size()
+}
+
```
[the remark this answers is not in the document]
ielashi: That can be done by making these methods `const`. I can take care of that separately.

src/memory_manager.rs
```diff
@@ -853,4 +1240,188 @@ mod test {
...
+    fn free_memory_that_was_not_used() {
+        let mut mem_mgr = MemoryManager::init(make_memory());
+        let memory_0 = mem_mgr.get(MemoryId(0));
+        assert_eq!(memory_0.grow(1), 0);
+
+        mem_mgr.free(MemoryId(5));
+    }
+
+    #[test]
+    fn upgrade_from_v1_to_v2() {
```
ielashi, line 1404: Is this enough? Upgrading from V1 to V2, if it has a bug, can lead to data loss, which is really bad. This test is testing a very simple use-case. We should ideally have a `proptest`, where we initialize a V1 memory manager with a random number of memories, write random data, then reload with v2, and verify that all the memories and the data exist. Have a look at `write_and_read_random_bytes` for inspiration.

src/memory_manager.rs
ielashi: I don't understand what you mean "consistent with nonconstant values", but it definitely can be a constant. I'll do it.

src/memory_manager.rs
ielashi: Just because something is mentioned in a research paper doesn't mean we should do it :) Using smaller bucket IDs for us allows to minimize our memory footprint compared to using larger bucket IDs.

src/memory_manager.rs
dragoljub-duric: Done.

src/memory_manager.rs
dragoljub-duric: Done.

src/memory_manager.rs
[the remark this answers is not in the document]
dragoljub-duric: I have tried that, but I have a problem since the `Address` implements `From<u64>`, and I am not sure how to make it return `const`.

src/memory_manager.rs
```diff
@@ -287,13 +393,99 @@ impl<M: Memory> MemoryManagerInner<M> {
...
+        let mem_mgr = if last_occupied_bucket == -1 {
+            Self {
+                memory,
+                allocated_buckets: header.num_allocated_buckets,
+                bucket_size_in_pages: header.bucket_size_in_pages,
+                memory_sizes_in_pages: header.memory_sizes_in_pages,
+                memory_buckets,
+                freed_buckets: BTreeSet::new(),
+            }
+        } else {
```
ielashi, line 440: I believe here is incorrect. Here's a test case that fails:

```
    #[test]
    fn freed_memories_are_tracked() {
        let mem = make_memory();
        let mut mem_mgr = MemoryManager::init(mem.clone());
        mem_mgr.get(MemoryId(0)).grow(1);
        mem_mgr.get(MemoryId(1)).grow(1);
        mem_mgr.free(MemoryId(0));

        let mem_mgr = MemoryManager::init(mem);
        assert_eq!(
            mem_mgr.inner.borrow().freed_buckets,
            maplit::btreeset! { BucketId(0) }
        );
    }
```

src/memory_manager.rs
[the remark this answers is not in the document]
ielashi: That can be addressed by creating a new `const` method on `Address`. We can take care of that later, it's not important.

src/memory_manager.rs
dragoljub-duric: Great point! I fixed it now, that was the problem with loading freed buckets when the Bucket ID is 0. PTAL.

src/memory_manager.rs
```diff
@@ -317,28 +509,32 @@ impl<M: Memory> MemoryManagerInner<M> {
...
-            let new_bucket_id = BucketId(self.allocated_buckets);
+            let new_bucket_id = match self.freed_buckets.pop_last() {
+                Some(t) => t,
+                None => {
+                    if self.allocated_buckets != MAX_NUM_BUCKETS as u16 {
+                        let new_id = self.allocated_buckets;
+                        self.allocated_buckets += 1;
+                        BucketId(new_id)
+                    } else {
+                        panic!("{id:?}: grow failed");
```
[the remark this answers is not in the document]
dragoljub-duric: Done

src/memory_manager.rs
```diff
@@ -317,28 +509,32 @@ impl<M: Memory> MemoryManagerInner<M> {
...
             return -1;
         }
 
         // Allocate new buckets as needed.
         for _ in 0..new_buckets_needed {
-            let new_bucket_id = BucketId(self.allocated_buckets);
+            let new_bucket_id = match self.freed_buckets.pop_last() {
+                Some(t) => t,
+                None => {
+                    if self.allocated_buckets != MAX_NUM_BUCKETS as u16 {
```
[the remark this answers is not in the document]
dragoljub-duric: Done

src/memory_manager.rs
```diff
@@ -317,28 +509,32 @@ impl<M: Memory> MemoryManagerInner<M> {
...
+            > MAX_NUM_BUCKETS
+        {
             // Exceeded the memory that can be managed.
             return -1;
         }
 
         // Allocate new buckets as needed.
         for _ in 0..new_buckets_needed {
-            let new_bucket_id = BucketId(self.allocated_buckets);
+            let new_bucket_id = match self.freed_buckets.pop_last() {
```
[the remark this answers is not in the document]
dragoljub-duric: That is right, it is changed now, and the test is added.

dsarlis:
@ielashi @dragoljub-duric Hey guys, this has been open for quite a while. Should we try to wrap it up?

ielashi:
> @ielashi @dragoljub-duric Hey guys, this has been open for quite a while. Should we try to wrap it up?

I'm now nearly done with the BTreeMap changes. I'll schedule time to wrap up the review of this tomorrow.

ielashi (commented):
In general I think the PR LGTM. I prefer we hold off on merging this though until we do the BTreeMap release (0.6.0), and then we can merge this change to release 0.6.1. This would be better so that we can minimize the changes in 0.6.0, which are already quite big.

ielashi:
> Do we want to include this change in the upcoming 0.6.0 release?

No, ideally this would follow in a minor release, so that we have had time to test thoroughly before releasing it.

dragoljub-duric:
@ielashi I saw that we have a new release. Can you approve it now?

ielashi:
> @ielashi I saw that we have a new release. Can you please approve it now?

Hey @dragoljub-duric, I'd like to do some more testing given that we found a number of bugs previously in the PR, and given how important this code is, it's better to err on the side of caution.

ielashi (commented):
Getting back to this PR now :) I left some early comments, still taking a closer look.

src/memory_manager.rs
```diff
@@ -287,13 +396,108 @@ impl<M: Memory> MemoryManagerInner<M> {
...
+
+        // Update the header.
+        mem_mgr.save_header();
+
+        // Rewrite the stable store to be layout V2.
+        write(
+            &mem_mgr.memory,
+            bucket_indexes_offset().get(),
+            mem_mgr.get_bucket_ids_in_bytes().as_ref(),
+        );
```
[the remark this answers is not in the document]
dragoljub-duric: I will need to give it a look again, I do not remember.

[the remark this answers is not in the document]
dragoljub-duric: The initial idea was to avoid saving when freeing the memory, so here it was saved, but that approach was changed, and this left. I will remove this part.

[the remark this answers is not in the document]
dragoljub-duric: However, this saving can be good to add when loading layout V1. That way we will be sure that the next time it will be loaded as layout V2 even if no changes to memory are made. WDYT?

src/memory_manager.rs
```diff
@@ -287,13 +397,96 @@ impl<M: Memory> MemoryManagerInner<M> {
             bucket_size_in_pages: header.bucket_size_in_pages,
             memory_sizes_in_pages: header.memory_sizes_in_pages,
             memory_buckets,
+            freed_buckets: BTreeSet::new(),
+        }
+    }
+
+    fn load_layout_v2(memory: M, header: Header) -> Self {
```
ielashi, line 404: I've been finding this function generally difficult to read. I'll add some pointers below on how to make it more readable.

src/memory_manager.rs
```diff
@@ -287,13 +397,96 @@ impl<M: Memory> MemoryManagerInner<M> {
...
+
+        const BYTE_SIZE_IN_BITS: usize = 8;
+        let buckets_index_size_in_bytes: usize =
+            (number_of_used_buckets as usize * BUCKET_ID_LEN_IN_BITS + (BYTE_SIZE_IN_BITS - 1))
+                / BYTE_SIZE_IN_BITS;
+
+        let mut buckets = vec![0; buckets_index_size_in_bytes];
+        memory.read(bucket_indexes_offset().get(), &mut buckets);
+
+        let buckets_decompressed = bytes_to_bucket_indexes(&buckets);
```
ielashi, line 423: One way to make code more easy to read is by scoping. For example, this block can be restructured to be like this:

```suggestion
        // Load the buckets.
        let buckets = {
            const BYTE_SIZE_IN_BITS: usize = 8;
            let buckets_index_size_in_bytes: usize =
                (number_of_used_buckets as usize * BUCKET_ID_LEN_IN_BITS + (BYTE_SIZE_IN_BITS - 1))
                    / BYTE_SIZE_IN_BITS;

            let mut buckets = vec![0; buckets_index_size_in_bytes];
            memory.read(bucket_indexes_offset().get(), &mut buckets);
            bytes_to_bucket_indexes(&buckets)
        };
```

In this case, it's clear that all this logic is trying to load buckets and, as a reader, I don't need to look inside this block unless I want to see how buckets are loaded.

src/memory_manager.rs
```diff
@@ -287,13 +397,96 @@ impl<M: Memory> MemoryManagerInner<M> {
...
+
+        let mut memory_buckets = BTreeMap::new();
+
+        let mut last_occupied_bucket: i32 = -1;
```
ielashi, line 427: Rather than having an `i32`, consider redefining this variable into something like this:

```suggestion
        // The last bucket that's accessed.
        let mut max_bucket_id: u16 = 0;
```

Not having `i32` will help get rid of the `if` statement below (see other comment)

src/memory_manager.rs
```diff
@@ -287,13 +397,96 @@ impl<M: Memory> MemoryManagerInner<M> {
...
+
+            Self {
+                memory,
+                allocated_buckets: header.num_allocated_buckets,
+                bucket_size_in_pages: header.bucket_size_in_pages,
+                memory_sizes_in_pages: header.memory_sizes_in_pages,
+                memory_buckets,
+                freed_buckets,
+            }
+        }
```
ielashi, line 472: Generally speaking, try to reduce if statements in your code. It results in more readable and maintainable code. Here's how I'd replace this if statement:


```
        let mut freed_buckets: BTreeSet<BucketId> = (0..max_bucket_id).map(BucketId).collect();

        for id in buckets.iter() {
            freed_buckets.remove(id);
        }

        Self {
            memory,
            allocated_buckets: header.num_allocated_buckets,
            bucket_size_in_pages: header.bucket_size_in_pages,
            memory_sizes_in_pages: header.memory_sizes_in_pages,
            memory_buckets,
            freed_buckets,
        }
```

dragoljub-duric: Good point, done.

src/memory_manager.rs
dragoljub-duric: Done. PTAL.

dragoljub-duric:
@ielashi Can you please take another look?

ielashi:
> @ielashi Can you please take another look?

@dragoljub-duric Thanks! From a code standpoint this looks good. The remaining concern I have is related to performance. Prior to this PR, memory performance operations were O(1), but with this change they would be O(# buckets used), correct? I don't yet know exactly how big of a deal this is, as we don't have benchmarks for that.

Let's discuss more in our 1:1 about next steps.

ielashi:
At a high-level, I think we should work on:

1. Verifying the correctness of this change.
  For that, I think we need stronger test-cases for the MemoryManager. I suggest we (in a separate PR), add a fuzz test, in a way that's similar to [this one](https://github.com/dfinity/stable-structures/blob/1224081801abd93e84d1859cd1241ed72b5a05f1/src/btreemap/proptests.rs#L72), and we can run this for a couple of days with these changes to ensure things are working as expected.

2. Understanding the performance impact of this change.
  * For that, I think we need to introduce (also in separate PR) the CI changes to track our benchmarks, just as we did with the `bitcoin-canister` repo.
  * We can then also add a benchmark where we allocate a large number of buckets and see how this current PR affects the performance of that benchmark.

ielashi:
We had a discussion offline and agreed that this approach may not be the best approach. This design was intended to keep the memory manager header bound to 64KiB to make migration from V1 simple, but that came with the additional complexity of using 15-bits for bucket indexes, and also a performance degradation where the `grow` memory operation can be up to 3 orders of magnitude more expensive.

In the future, we’ll likely need to expand the memory manager header beyond 64KiB in order to, for example, support a larger number of buckets. If that’s going to happen eventually, then there are simpler and more efficient ways to implement memory freeing. For example, we won’t need to store the bucket IDs as 15-bits, and we can use linked lists to make growing and freeing at O(# buckets), as opposed to O(max number of buckets) with the current approach.

There hasn’t been a lot of pressure on us to deliver this feature, so we can park this for now, close the ticket, and revisit with a more comprehensive design in the future.

PR #99, opened 2023-07-06
chore: add Hash derive for MemoryId

witter-deland:


src/memory_manager.rs
```diff
@@ -518,6 +518,30 @@ impl MemoryId {
...
+    const IS_FIXED_SIZE: bool = true;
+}
+
+impl std::ops::Deref for MemoryId {
+    type Target = u8;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
roman-kashitsyn, line 543: Is there a real need to treat memory identifiers as integers?
I'd avoid allowing conversions from MemoryId and u8 unless there is a pressing need.

src/memory_manager.rs
```diff
@@ -518,6 +518,30 @@ impl MemoryId {
...
+    fn from_bytes(bytes: std::borrow::Cow<[u8]>) -> Self {
+        Self(u8::from_be_bytes(bytes.as_ref().try_into().unwrap()))
+    }
+}
+
+impl crate::BoundedStorable for MemoryId
+{
+    const MAX_SIZE: u32 = 1;
+    const IS_FIXED_SIZE: bool = true;
+}
```
roman-kashitsyn, line 535: Note that this code will complicate supporting more than 255 in the future.
Here, we must commit to using 8 bits for the memory identifier.

roman-kashitsyn (approved):
I'm somewhat worried about this change because it requires the library to commit to a specific MemoryId representation which might be hard to change in the future without breaking all the clients.

Apriory, I'd suggest using `u8` constants in your code and converting them to MemoryId upon decoding.

I'd like to hear more about your use case.

ielashi:
> I'm somewhat worried about this change because it requires the library to commit to a specific MemoryId representation which might be hard to change in the future without breaking all the clients.
> 
> Apriory, I'd suggest using `u8` constants in your code and converting them to MemoryId upon decoding.
> 
> I'd like to hear more about your use case.

I have a similar concern. I think it's quite likely that we'd want to support > 255 memories in the future, and this change would complicate that transition. I'd also like to hear more about your use-case.

witter-deland:
I'm trying to provide a wrapper for Memory to detect duplicate memory id usage when many stable structures are used in the canister. However, based on the situation you describe, I think there are other alternative ways to meet my needs. Let me adjust the code...
```rust
use std::cell::RefCell;
use std::collections::HashSet;
use std::ops::Deref;

use ic_stable_structures::memory_manager::{MemoryId, MemoryManager, VirtualMemory};
use ic_stable_structures::DefaultMemoryImpl;

pub type Memory = VirtualMemory<DefaultMemoryImpl>;

thread_local! {
    static MEMORY_MANAGER: RefCell<MemoryManager<DefaultMemoryImpl>> =
        RefCell::new(MemoryManager::init(DefaultMemoryImpl::default()));
    static MEMORY_ID_ALLOCATED: RefCell<HashSet<u8>> = RefCell::new(HashSet::new());
}

pub fn get_memory(id: MemoryId) -> Memory {
    assert!(
        !MEMORY_ID_ALLOCATED.with(|m| m.borrow().contains(id.deref())),
        "duplicate memory id"
    );
    MEMORY_ID_ALLOCATED.with(|m| m.borrow_mut().insert(id.deref().clone()));
    MEMORY_MANAGER.with(|m| m.borrow().get(id))
}
```
> I'm somewhat worried about this change because it requires the library to commit to a specific MemoryId representation which might be hard to change in the future without breaking all the clients.
> 
> Apriory, I'd suggest using `u8` constants in your code and converting them to MemoryId upon decoding.
> 
> I'd like to hear more about your use case.

witter-deland:
I would change the code to something like this:
```rust
static MEMORY_ID_ALLOCATED: RefCell<HashSet<MemoryId>> = RefCell::new(HashSet::new())
```

There is no way to convert MemoryId into a number, and providing a Hash derive can meet my needs. Of course, it is possible to use Vec without adding Hash derive... The algorithm complexity performance of HashSet O(1) will be better, so I think adding Hash derive is a better choice

> I'm trying to provide a wrapper for Memory to detect duplicate memory id usage when many stable structures are used in the canister. However, based on the situation you describe, I think there are other alternative ways to meet my needs. Let me adjust the code...
> 
> ```rust
> use std::cell::RefCell;
> use std::collections::HashSet;
> use std::ops::Deref;
> 
> use ic_stable_structures::memory_manager::{MemoryId, MemoryManager, VirtualMemory};
> use ic_stable_structures::DefaultMemoryImpl;
> 
> pub type Memory = VirtualMemory<DefaultMemoryImpl>;
> 
> thread_local! {
>     static MEMORY_MANAGER: RefCell<MemoryManager<DefaultMemoryImpl>> =
>         RefCell::new(MemoryManager::init(DefaultMemoryImpl::default()));
>     static MEMORY_ID_ALLOCATED: RefCell<HashSet<u8>> = RefCell::new(HashSet::new());
> }
> 
> pub fn get_memory(id: MemoryId) -> Memory {
>     assert!(
>         !MEMORY_ID_ALLOCATED.with(|m| m.borrow().contains(id.deref())),
>         "duplicate memory id"
>     );
>     MEMORY_ID_ALLOCATED.with(|m| m.borrow_mut().insert(id.deref().clone()));
>     MEMORY_MANAGER.with(|m| m.borrow().get(id))
> }
> ```
> 
> > I'm somewhat worried about this change because it requires the library to commit to a specific MemoryId representation which might be hard to change in the future without breaking all the clients.
> > Apriory, I'd suggest using `u8` constants in your code and converting them to MemoryId upon decoding.
> > I'd like to hear more about your use case.

ielashi:
@witter-deland Questions:

* I assume from the code sample you shared that you are persisting `MEMORY_ID_ALLOCATED` at some point in your `pre_upgrade` hook?
* Are you allocating new stable structures at run-time? I would generally recommend against that for now. If you're allocating all your stable structures at compile-time, then there are easier ways to prevent a memory ID collision than the code you've shared.

roman-kashitsyn (approved):
LGTM!

witter-deland:
No, it will not save.

> * assume from the code sample you shared that you are persisting `MEMORY_ID_ALLOCATED` at some point in your `pre_upgrade` hook?

No, I'm not allocating at runtime. This code is only to assist in the detection of MemoryId reuse during the testing phase, because once there is reuse, this detection mechanism will trigger an assert when Canister is running. This can help developers avoid MemoryId reuse during the testing phase.

And, would you like to share the easier way you say it?
> * Are you allocating new stable structures at run-time? I would generally recommend against that for now. If you're allocating all your stable structures at compile-time, then there are easier ways to prevent a memory ID collision than the code you've shared

ielashi:
@witter-deland I see. I was under the impression that you were allocating stable structures at run-time, and thus you would need to persist the HashMap. OK, given that, the change LGTM.

> And, would you like to share the easier way you say it?

In the Bitcoin Canister, for instance, the way we avoid reuse of the same memory ID is by:

1. Keeping the `MemoryManager` private to one particular file ([memory.rs](https://github.com/dfinity/bitcoin-canister/blob/master/canister/src/memory.rs))
2. Not exposing a method where a caller can request an arbitrary `MemoryId`. Instead, a method is exposed to retrieve each memory explicitly (`get_balances_memory`, `get_upgrades_memory`, etc.)
3. All the memories used are hard-coded as constants [here](https://github.com/dfinity/bitcoin-canister/blob/master/canister/src/memory.rs#L13) in sorted order, so it's quite easy to see which memory IDs are already being used.

This doesn't give the same runtime-check that you'd like to do, but we found it sufficient for our purposes. Nonetheless, I'm happy to merge your PR if you'd like to continue with your current solution.

witter-deland:
Thank you for your kind reply.

This way has better code maintainability, I choose to use your way. This PR can be closed (no need to merge).

> In the Bitcoin Canister, for instance, the way we avoid reuse of the same memory ID is by:
> 
> 1. Keeping the `MemoryManager` private to one particular file ([memory.rs](https://github.com/dfinity/bitcoin-canister/blob/master/canister/src/memory.rs))
> 2. Not exposing a method where a caller can request an arbitrary `MemoryId`. Instead, a method is exposed to retrieve each memory explicitly (`get_balances_memory`, `get_upgrades_memory`, etc.)
> 3. All the memories used are hard-coded as constants [here](https://github.com/dfinity/bitcoin-canister/blob/master/canister/src/memory.rs#L13) in sorted order, so it's quite easy to see which memory IDs are already being used.

Issue #98, opened 2023-07-05
vec::InitError is now private


veeso:
Since 0.5.5 `vec::InitError` is private, which means the error returned by Vec::init cannot be accessed. Are there any reasons behind this or is it a mistake?

ielashi:
@veeso Thanks for reporting this. I believe this is a mistake. We'll fix this and publish a new release shortly.

ielashi:
@veeso We've now released 0.5.6, which should address this problem. Can you please try upgrading and let us know if everything is working the way you expect?

veeso:
Thank you for the quick response! It's accessible now

PR #100, opened 2023-07-06
fix: make `vec::InitError` public again

ielashi:
The refactoring of `vec` accidentally made `vec::InitError` private, which broke clients that upgraded to the latest stable structures release. See issue #98.

This commit makes `vec::InitError` public, and adds a test to ensure it always remains public.


roman-kashitsyn (approved):
LGTM, thank you!

PR #105, opened 2023-07-19
POC: A dynamic memory allocator for BTreeMap

ielashi:
This PR introduces a dynamic memory allocator implementation, which will be used in the BTreeMap to allow for unbounded types. The allocator uses a two-level segregated free list to manage allocations, which is described in [this paper](http://www.gii.upv.es/tlsf/files/ecrts04_tlsf.pdf) by Masmano et al.

The implementation is functionally complete, but there are a few missing pieces:
1. Logic for upgrading from the BTreeMap's current constant-size allocator.
2. More testing, better documentation, and some minor cleanups.

- `tlsf.rs`
- `tlsf/blocks.rs`
- `tlfs/free_lists.rs`

This PR will be broken down into smaller bite-sized PRs.

While this allocator will initially only be used by BTreeMap, it is designed to be general-purpose and can be used by other structures in the future.


ielashi:
Closing this PR as it's already being broken up into smaller PRs (see #94, #114, and more on the way).

ielashi:
Closing this PR as we ended up using a different approach for BTreeMap that relies on fixed-size pages. The reason we opted for the latter solution rather than a dynamic memory allocator is that it's safer. A dynamic memory allocator may be more efficient, but depending on the usage pattern the amount of external fragmentation is unbounded.

We many potentially revisit this idea in the future though.

PR #107, opened 2023-07-21
test: `Node::load` and `Node::save`

ielashi:
A new version of `Node` will soon be introduced, and this commit paves the way for that by:

1. Moving the logic of `load` and `store` into a separate file (`v1.rs`)
2. Adding a proptest for `load` and `store` to ensure we don't introduce regressions

### DIFF
--- a/src/btreemap/node/tests.rs
+++ b/src/btreemap/node/tests.rs
@@ -1,0 +1,39 @@
+use super::*;
+use crate::types::NULL;
+use proptest::collection::btree_map as pmap;
+use proptest::collection::vec as pvec;
+use std::cell::RefCell;
+use std::collections::BTreeMap;
+use std::rc::Rc;
+use test_strategy::proptest;
+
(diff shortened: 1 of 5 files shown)


src/btreemap/node/tests.rs
```diff
@@ -0,0 +1,39 @@
...
+use std::cell::RefCell;
+use std::collections::BTreeMap;
+use std::rc::Rc;
+use test_strategy::proptest;
+
+fn make_memory() -> Rc<RefCell<Vec<u8>>> {
+    Rc::new(RefCell::new(Vec::new()))
+}
+
+#[proptest]
```
ielashi, line 14: In the original `proptest!` macro, code formatting doesn't work and has to be formatted manually, which I always find annoying. This attribute macro from the `test_strategy` crate is a wrapper around `proptest`, but it has the advantage that code formatting still works, along with other syntax simplifications.

roman-kashitsyn (approved):

PR #108, opened 2023-07-21
chore: update ic-cdk to 0.10

lwshang:


Cargo.toml
```diff
@@ -11,10 +11,12 @@ keywords = ["internet-computer", "dfinity", "stable-structures"]
 include = ["src", "Cargo.toml", "LICENSE", "README.md"]
 repository = "https://github.com/dfinity/stable-structures"
 
+[dependencies]
+ic0 = "0.18.11"
```
ielashi, line 15: We try our best to keep this library dependency-free. Is there value in using `ic0` here?

ielashi (commented):
Is there some motivational reason why we're making this upgrade, or is the PR only with the intention of keeping dependencies up to date?

Cargo.toml
chenyan-dfinity: This is mainly to replace the imports in [ic0_memory.rs](https://github.com/dfinity/stable-structures/pull/108/files#diff-67ba0429f16d00d3051dba3557d7d97485a419938486a684de2892c3fe735065). `ic0` merely copies the API from the spec, which this library implicitly depends on anyway.

One benefit of using `ic0` is that we have a `wasi` feature flag that turns off the ic imports, which we use to generate did file from CDK. If we don't include this dependency, we get errors like this: https://dfinity.slack.com/archives/C04L102C80Z/p1691572642587119

ielashi: This is new to me so, having briefly looked into it, my understanding is that:

* `ic-cdk` provides a way to generate the candid file automatically from a Rust canister.
* The generation of the candid depends on the feature `ic-cdk/wasi` to remove `ic0` imports.

IIUC, in order for this to work, every single library that a Rust canister depends on must meet one of these two conditions:
* It isn't using the system API at all.
* It's using `ic-cdk`, or at least depending on the `ic-cdk/ic0` crate because of the hidden `ic-cdk/wasi` feature.

With this approach, are we now forcing all library developers to depend on `ic-cdk`, and any other library would be practically unusable, correct? Instead of relying on a feature like `ic-cdk/wasi`, which is specific to `ic-cdk`, would it not be possible to provide stubs for the ic0 API in the process of candid generation? That would work universally and wouldn't require libraries to depend on the `ic-cdk/wasi` feature.

> This is mainly to replace the imports in [ic0_memory.rs](https://github.com/dfinity/stable-structures/pull/108/files#diff-67ba0429f16d00d3051dba3557d7d97485a419938486a684de2892c3fe735065). ic0 merely copies the API from the spec, which this library implicitly depends on anyway.

I'm not sure I agree with that. `stable-structures` of course depends on the system API, but what we're really depending on here is the `ic-cdk/wasi` feature, which is an `ic-cdk` concept and is orthogonal to the system API itself.

cc @peterpeterparker since he's also encountering the issue. #110 is also a report of something similar.

lwshang: I believe the root cause is that the crate features `ic0/wasi` and `ic-cdk/wasi` are not additive.

Maybe we can mock the ic0 system API in `wasmtime` so that the same wasm module can run in both canister execution environment and `wasmtime`. And we won't need the conditional compilation.

ielashi: @lwshang @chenyan-dfinity So is this the conclusion:

> Maybe we can mock the ic0 system API in wasmtime so that the same wasm module can run in both canister execution environment and wasmtime. And we won't need the conditional compilation.

If yes, that means this PR can be closed, right?

lwshang: Yes. I've got a POC for it.
It will take some time to be integrated in `dfx` and standardized as a common approach for all language CDKs.

chenyan-dfinity: Not necessarily a mock, `ic-wasm` can do another pass to remove unknown import modules.

PR #113, opened 2023-08-15
test: improve test coverage of `BTreeMap::Node`

ielashi:
Makes the proptest for nodes more rigorous by:

* Generating internal nodes that include children, and not just leaf nodes.
* Generating nodes with zero-sized keys and values

Enhancements to the documentation are also included.


src/btreemap/node/v1.rs
```diff
@@ -131,3 +187,17 @@ impl<K: Storable + Ord + Clone> Node<K> {
         }
     }
 }
+
+/// Returns the size of a v1 node in bytes.
+pub(super) fn size_v1(max_key_size: u32, max_value_size: u32) -> Bytes {
+    let max_key_size = Bytes::from(max_key_size);
+    let max_value_size = Bytes::from(max_value_size);
```
dsarlis, line 194: nit: I'd move this below, just before we compute `entry_size` so the definition is closer where it's used.

dsarlis (approved):

PR #114, opened 2023-08-15
feat: v2 of `BTreeMap::Node` to support unbounded types.

ielashi:
Introduces V2 of `BTreeMap::Node`, which has a smaller memory footprint than V1 and includes support for unbounded types.

The implementation is currently not very efficient, as it deserializes all entries on load and serializes all entries on save. Performance enhancements will be included in subsequent PRs.

Note that V2 here is only used in tests and production continues to use V1.


src/btreemap/node/v2.rs
```diff
@@ -0,0 +1,402 @@
...
+//! ----------------------------------------
+//! Data
+//! ----------------------------------------
+//! ```
+//!
+//! ## Keys and Values
+//! Keys and values are both encoded in memory as blobs.
+//!
+//! If they are variable in size (i.e. their `IS_FIXED` attribute is set to false),
+//! then the size of the blob is encoded before the blob itself. Otherwise, no size
```
dsarlis, line 67: Should we be explicit that the size is encoded as a `u32` to set expectations?

src/btreemap/node/v2.rs
```diff
@@ -0,0 +1,402 @@
...
+
+        // Append its data to the buffer.
+        buf.extend_from_slice(&overflow_buf[PAGE_OVERFLOW_DATA_OFFSET.get() as usize..]);
+    }
+
+    buf
+}
+
+fn read_u16_from_slice(slice: &[u8], offset: usize) -> u16 {
+    u16::from_le_bytes((&slice[offset..offset + 2]).try_into().unwrap())
```
dsarlis, line 385: Perhaps we use `expect` here to be able to provide a somewhat better message?

src/btreemap/node/v2.rs
```diff
@@ -0,0 +1,402 @@
...
+        buf.extend_from_slice(&[LAYOUT_VERSION_2]);
+        buf.push(match self.node_type {
+            NodeType::Leaf => LEAF_NODE_TYPE,
+            NodeType::Internal => INTERNAL_NODE_TYPE,
+        });
+        buf.extend_from_slice(&(self.keys.len() as u16).to_le_bytes());
+
+        // Add a null overflow address. This might get overwritten later in case the node
+        // does overflow.
+        buf.extend_from_slice(&[0; 8]);
```
dsarlis, line 211: It's probably a low impact change and perhaps you plan to address in follow ups anyway but I'll still ask: can we create a vector with capacity for the first few things we need to add to the buffer? It seems like it should be possible because we should know how much space these take. (The keys and values later need to be done on demand of course).

src/btreemap/node/v2.rs
```diff
@@ -0,0 +1,402 @@
...
+    // Pages can be allocated and deallocated as needed.
+    fn write_paginated<M: Memory>(
+        &self,
+        buf: Vec<u8>,
+        allocator: &mut Allocator<M>,
+        page_size: usize,
+    ) {
+        // Compute how many overflow pages are needed.
+        let additional_pages_needed = if buf.len() > page_size {
+            let overflow_page_capacity = page_size - PAGE_OVERFLOW_DATA_OFFSET.get() as usize;
```
dsarlis, line 258: The subtraction here I suppose assumes that `page_size` > PAGE_OVERFLOW_DATA_OFFSET which should be true indeed. Perhaps we throw an `assert!` here about this condition just to be extra careful?

src/btreemap/node/v2.rs
```diff
@@ -0,0 +1,402 @@
...
+
+    // Append overflow pages, if any.
+    let mut overflow_address = address_from_slice(&buf, OVERFLOW_ADDRESS_OFFSET);
+    while overflow_address != NULL {
+        // Read the overflow.
+        let mut overflow_buf = vec![0; page_size as usize];
+        memory.read(overflow_address.get(), &mut overflow_buf);
+
+        // Validate the magic of the overflow.
+        assert_eq!(&overflow_buf[0..3], OVERFLOW_MAGIC, "Bad magic.");
```
dsarlis, line 372:
```suggestion
        assert_eq!(&overflow_buf[0..3], OVERFLOW_MAGIC, "Bad overflow magic.");
```

src/btreemap/node/v2.rs
ielashi: Done.

src/btreemap/node.rs
```diff
@@ -413,3 +424,61 @@ enum Value {
...
+    V1 {
+        max_key_size: u32,
+        max_value_size: u32,
+    },
+    /// V2 nodes have a fixed page size.
+    V2(PageSize),
+}
+
+impl Version {
+    fn page_size(&self) -> u32 {
```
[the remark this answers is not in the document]
dsarlis: Gotcha. In that case maybe we need to find a way to encode that V1 also has a `PageSize` concept. I think the disconnect for me was that in the `Version` type, `V1` does not have an associated `PageSize` like `V2` does. Maybe we actually use a `PageSize` there as well (I suppose it would be the variant `PageSize::Derived` in that case). Not entirely sure this makes sense or is indeed better but thought I'd throw it out there.

src/btreemap/node/v2.rs
dsarlis: If not mistaken the assertion you mention is `page_size >= MINIMUM_PAGE_SIZE`. I suppose this should be enough given that `MINIMUM_PAGE_SIZE` > `PAGE_OVERFLOW_DATA_OFFSET` but was thinking that an explicit `page_size >= PAGE_OVERFLOW_DATA_OFFSET` doesn't hurt so we might as well add it.

dsarlis (approved):

src/btreemap/node.rs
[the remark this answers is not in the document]
ielashi: I see your point, and it does makes sense. I introduced a `DerivedPageSize` type to make this more clear per your suggestion.

src/btreemap/node/v2.rs
ielashi: Sure, done.

PR #121, opened 2023-09-01
feat: introduce new benchmarking script

ielashi:
## Problem
We've been relying on `criterion` as our way of running benchmarks and getting reports. However, `criterion` has been falling short for a number of reasons:

1. Benchmarks can only have one measurement. In our case, the measurement was "instructions", but that's not good enough. For our use-case, we'd like to have multiple measurements per benchmark (e.g. instructions, stable memory pages, etc.).
2. `criterion` always compares the results to the previous run. In almost all cases, I'm interested in measuring the difference in performance relative to the last commit, and there isn't a good of enforcing that workflow in `criteron`.

## Solution
I decided to spend a few hours to replace our reliance on `criterion` altogether. We now have our very own benchmarking script that runs the benchmarks. A benchmark can have as many measurements as we want, and it neatly persists these benchmarks to our repo along with a report on how the benchmarks have changed.

Here's an example of the report that it prints out to the terminal:
image

Notes:
* As you can tell from the screen shot above, I added the "stable_memory_size" as a new measurement.
* I had to bump up the rust version as `clap` needed a more rust >= 1.70.


Cargo.toml
```diff
@@ -11,26 +11,41 @@ keywords = ["internet-computer", "dfinity", "stable-structures"]
 include = ["src", "Cargo.toml", "LICENSE", "README.md"]
 repository = "https://github.com/dfinity/stable-structures"
 
+# There are no benchmarks in the library itself, so by default it's turned off
+# to avoid having `cargo bench` looks for benchmarks there.
```
dsarlis, line 15:
```suggestion
# to avoid having `cargo bench` look for benchmarks there.
```

benchmarks/benchmark.rs
```diff
@@ -0,0 +1,260 @@
+//! A script for running benchmarks on a canister.
```
dsarlis, line 1: nit: Should we add an example usage of the script (e.g. with the arguments that we need in the average case).

dsarlis (approved):

benchmarks/benchmark.rs
ielashi: Done. It's just `cargo bench` by the way.

PR #123, opened 2023-09-11
perf: update BTreeMap v2 default page size from 500 bytes to 1024 bytes

ielashi:
The page size used by the BTreeMap dictates how big the chunks are that are allocated. The smaller the page size, the more chunks that need to be allocated to store the same amount of data, and consequently the more reads and writes are required to load/save each node. On the other hand, a large page size also can incur additional memory overhead, so we need to be careful to not set it too high.

To explore this, I added a benchmark that stores 200 10MiB vectors in a BTreeMap. I tested different page sizes to explore its effect on both memory and instructions.

When using a larger pages size, from 1KiB to 20MiB, this is the best increase in performance I had seen:

(relative performance is measured against the base line of `page_size = 500`)
```
Benchmark: btreemap_insert_10mib_values
  measurements:
    instructions: 252642984 (improved by 16.29%)
    node_load_v2: 3607078 (improved by 76.06%)
    node_save_v2: 227974389 (improved by 13.72%)
    stable_memory_size: 596 (regressed by 1652.94%)
```

And I tried a page size of 1024:

```
Benchmark: btreemap_insert_10mib_values
  measurements:
    instructions: 278228334 (improved by 7.81%)
    node_load_v2: 10377468 (improved by 31.12%)
    node_save_v2: 245958554 (improved by 6.92%)
    stable_memory_size: 33 (improved by 2.94%)
```

As we can see, we captured nearly half of the performance improvement without increasing the page size too much, which seems to strike a good balance and it makes sense to have it as a default value.

### DIFF
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -46,10 +46,13 @@ mod proptests;
 const MAGIC: &[u8; 3] = b"BTR";
 const LAYOUT_VERSION: u8 = 1;
 const LAYOUT_VERSION_2: u8 = 2;
-/// The sum of all the header fields, i.e. size of a packed header.
+// The sum of all the header fields, i.e. size of a packed header.
 const PACKED_HEADER_SIZE: usize = 28;
-/// The offset where the allocator begins.
+// The offset where the allocator begins.
 const ALLOCATOR_OFFSET: usize = 52;
+
+// The default page size to use in BTreeMap V2.
+const DEFAULT_PAGE_SIZE: u32 = 1024;
 
 /// A "stable" map based on a B-tree.
 ///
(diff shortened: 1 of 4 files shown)


src/btreemap.rs
```diff
@@ -46,11 +46,14 @@ mod proptests;
...
 const LAYOUT_VERSION: u8 = 1;
 const LAYOUT_VERSION_2: u8 = 2;
-/// The sum of all the header fields, i.e. size of a packed header.
+// The sum of all the header fields, i.e. size of a packed header.
 const PACKED_HEADER_SIZE: usize = 28;
-/// The offset where the allocator begins.
+// The offset where the allocator begins.
 const ALLOCATOR_OFFSET: usize = 52;
 
+// The default page size to use in BTreeMap V2.
```
dsarlis, line 54:
```suggestion
// The default page size to use in BTreeMap V2 in bytes.
```

dsarlis (approved):

PR #125, opened 2023-09-12
bug: contains_key doesn't work in btree v2

witter-deland:
cargo test insert_stable_principal

### DIFF
--- a/src/btreemap/proptests.rs
+++ b/src/btreemap/proptests.rs
@@ -91,6 +91,8 @@ fn insert_stable_principal() {
         assert_eq!(key1, key2);
 
         btree.insert(key1.clone(), ());
+
+        assert_eq!(btree.len(), 1);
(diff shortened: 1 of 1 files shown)


src/btreemap/proptests.rs
```diff
@@ -62,3 +64,35 @@ proptest! {
...
+    fn to_bytes(&self) -> std::borrow::Cow<[u8]> {
+        self.0.as_ref().into()
+    }
+
+    fn from_bytes(bytes: std::borrow::Cow<[u8]>) -> Self {
+        Self(Principal::from_slice(bytes.as_ref()))
+    }
+    const BOUND: Bound = Bound::Bounded {
+        max_size: 29,
+        is_fixed_size: true,
```
roman-kashitsyn, line 81: Principals aren't fixed-size
```suggestion
        is_fixed_size: false,
```

PR #128, opened 2023-09-13
feat: impl Storable for Principal

witter-deland:


roman-kashitsyn:
Note that `candid` is a dev dependency, not the library dependency, so CI fails with the following error:

```
 --> src/storable.rs:6:5
  |
6 | use candid::Principal;
  |     ^^^^^^ use of undeclared crate or module `candid`

```
Given how hard it is to bump Candid versions, we'd prefer not to add `candid` as a dependency to add a few `impls`.

witter-deland:
Got it.  
> Note that `candid` is a dev dependency, not the library dependency, so CI fails with the following error:
> 
> ```
>  --> src/storable.rs:6:5
>   |
> 6 | use candid::Principal;
>   |     ^^^^^^ use of undeclared crate or module `candid`
> ```

> 
> Given how hard it is to bump Candid versions, we'd prefer not to add `candid` as a dependency to add a few `impls`.

witter-deland:
One personal opinion is: the types in candid are unavoidable in canister development, especially when stored in a stable structure. Supporting these types will be more friendly to developers.

roman-kashitsyn:
> the types in candid are unavoidable in canister development

I agree, but we don't want to force the specific Candid version on the users. We went through a very painful 0.8 -> 0.9 candid upgrade recently; the fewer libraries fix a specific candid version, the easier it's to coordinate version bumps.

witter-deland:
> I agree, but we don't want to force the specific Candid version on the users. We went through a very painful 0.8 -> 0.9 candid upgrade recently; the fewer libraries fix a specific candid version, the easier it's to coordinate version bumps.

Got it, thank you for your reply, this PR can be closed

PR #129, opened 2023-09-13
feat: impl Storable for Nat

witter-deland:


witter-deland:
Same reason like #128

PR #132, opened 2023-09-18
feat(log): add an iterator over thread-local log

roman-kashitsyn:
This change introduces a stable log iterator that can enumerate entries in stable log stored in a thread-local variable.

This simplifies dealing with stable logs a lot; currently, clients have to replicate the same logic for each log, see ckBTC for example[^1].

[^1]: https://github.com/dfinity/ic/blob/712910487fe14924cb80622af47b31210cab95bc/rs/bitcoin/ckbtc/minter/src/storage.rs#L37

### DIFF
--- a/src/log.rs
+++ b/src/log.rs
@@ -425,3 +425,66 @@ where
         self.next()
     }
 }
+
+/// Returns an iterator over entries in the log stored in a thread-local variable.
+pub fn iter_thread_local<T, I, D>(
+    local_key: &'static std::thread::LocalKey<std::cell::RefCell<Log<T, I, D>>>,
+) -> ThreadLocalRefIterator<T, I, D>
+where
+    T: Storable,
+    I: Memory,
+    D: Memory,
+{
+    ThreadLocalRefIterator {
+        log: local_key,
+        buf: vec![],
+        pos: 0,
+    }
+}
+
+pub struct ThreadLocalRefIterator<T, I, D>
+where
+    T: Storable + 'static,
+    I: Memory + 'static,
+    D: Memory + 'static,
+{
+    log: &'static std::thread::LocalKey<std::cell::RefCell<Log<T, I, D>>>,
+    buf: Vec<u8>,
+    pos: u64,
+}
+
+impl<T, I, D> Iterator for ThreadLocalRefIterator<T, I, D>
+where
+    T: Storable,
+    I: Memory,
+    D: Memory,
+{
+    type Item = T;
+
+    fn next(&mut self) -> Option<T> {
+        self.log.with(
+            |log| match log.borrow().read_entry(self.pos, &mut self.buf) {
+                Ok(()) => {
+                    self.pos = self.pos.saturating_add(1);
+                    Some(T::from_bytes(Cow::Borrowed(&self.buf)))
+                }
+                Err(NoSuchEntry) => None,
+            },
+        )
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let count = self.log.with(|cell| cell.borrow().len());
+        (count.saturating_sub(self.pos) as usize, None)
+    }
+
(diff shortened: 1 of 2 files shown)


src/log.rs
```diff
@@ -425,3 +425,66 @@ where
         self.next()
     }
 }
+
+/// Returns an iterator over entries in the log stored in a thread-local variable.
+pub fn iter_thread_local<T, I, D>(
+    local_key: &'static std::thread::LocalKey<std::cell::RefCell<Log<T, I, D>>>,
```
maksymar, line 431: nit: import repeated types

```rs
use std::thread::LocalKey;
use std::cell::RefCell;
```

src/log.rs
```diff
@@ -425,3 +425,66 @@ where
...
+                    self.pos = self.pos.saturating_add(1);
+                    Some(T::from_bytes(Cow::Borrowed(&self.buf)))
+                }
+                Err(NoSuchEntry) => None,
+            },
+        )
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let count = self.log.with(|cell| cell.borrow().len());
```
maksymar, line 477: I suppose this is the same as in count(), so can be simplified

```rs
fn size_hint(&self) -> (usize, Option<usize>) {
  (self.count(), None)
}
```

roman-kashitsyn: Unfortunately, that's impossible because `count()` consumes the iterator. I implemented it the other way around.

src/log.rs
```diff
@@ -425,3 +425,66 @@ where
...
+    ThreadLocalRefIterator {
+        log: local_key,
+        buf: vec![],
+        pos: 0,
+    }
+}
+
+pub struct ThreadLocalRefIterator<T, I, D>
+where
+    T: Storable + 'static,
```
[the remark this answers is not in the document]
roman-kashitsyn: IMO, there is no need for non-static thread-local keys (at least I've never used them). It's not even clear to me why anyone would want to use non-global thread locals.

src/log.rs
roman-kashitsyn: Done.

src/log.rs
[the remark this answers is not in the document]
maksymar: SGTM

maksymar (approved):

src/log.rs
```diff
@@ -425,3 +427,65 @@ where
         self.next()
     }
 }
+
+/// Returns an iterator over entries in the log stored in a thread-local variable.
+pub fn iter_thread_local<T, I, D>(
+    local_key: &'static LocalKey<RefCell<Log<T, I, D>>>,
```
ielashi, line 433: Is there a reason why we're coupling the concept of thread locals with the functionality of iteration, as opposed to adding a `.iter()` method on the `Log` structure directly as we did with `BTreeMap`.

I get that it's reasonable to assume that logs would typically be stored in thread locals, but that's not always the case. For example, in the Bitcoin canister we store stable structures inside structs and not directly as global variables ([example](https://github.com/dfinity/bitcoin-canister/blob/4b48647273f66b1832239e6dc662c317208551d0/canister/src/block_header_store.rs#L15)).

roman-kashitsyn: The Log structure already has `iter()`; my problem is that I want to hide the fact that the log lives in a stable structure stored in a thread-local refcell from the rest of the code.

I want a module interface where the client can `record_event` and get an event iterator: https://sourcegraph.com/github.com/dfinity/ic@595c57c815aebc49926a91618e2cde1cb7454973/-/blob/rs/bitcoin/ckbtc/minter/src/storage.rs?L75-96

That's not possible without a custom iterator that looks up values in the thread-local.

ielashi: I see, thanks for the clarification.

ielashi (approved):

PR #133, opened 2023-09-19
refactor: determine node version automatically on load

ielashi:
Prior to this commit, the caller of `Node::load` needed to specify the node version that will be loaded. This is acceptable if all nodes have the same version, but given that a BTreeMap may have different versions of nodes (migration will be introduced in subsequent commits), this needs to change such that the node's version is detected on load.

### DIFF
--- a/src/btreemap/node.rs
+++ b/src/btreemap/node.rs
@@ -21,7 +21,8 @@ use io::NodeReader;
 const B: usize = 6;
 // The maximum number of entries per node.
 const CAPACITY: usize = 2 * B - 1;
-const LAYOUT_VERSION: u8 = 1;
+const LAYOUT_VERSION_1: u8 = 1;
+const LAYOUT_VERSION_2: u8 = 2;
 const MAGIC: &[u8; 3] = b"BTN";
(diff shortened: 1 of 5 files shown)


src/btreemap/node.rs
```diff
@@ -65,13 +66,24 @@ pub struct Node<K: Storable + Ord + Clone> {
...
+            LAYOUT_VERSION_1 => {
+                if let PageSize::Derived(DerivedPageSize {
+                    max_key_size,
+                    max_value_size,
+                }) = page_size
+                {
+                    Self::load_v1(header, address, max_key_size, max_value_size, memory)
+                } else {
+                    unreachable!("Tried to load a V1 node without a derived PageSize.")
+                }
```
dsarlis, line 83: nit: Use a match since we want to handle both cases? It's a bit more idiomatic and we prevent accidentally not handling some case if we add more in the future.

dsarlis (approved):

src/btreemap/node.rs
ielashi: Done.

PR #134, opened 2023-09-20
feat: add additional assertions to the `Storable` trait.

ielashi:
This commit adds a new method, `to_bytes_checked`, to the `Storable` trait. It includes assertions to detect if the `Storable` implementation of an element is buggy. This addresses the issue in #125, where a user had a mistake in their implementation of Storable, but it wasn't clear from the error that this was the case.

BTreeMap has been updated to use `to_bytes_checked` rather than `to_bytes`.

### DIFF
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -300,34 +300,7 @@ where
     ///   key.to_bytes().len() <= max_size(Key)
     ///   value.to_bytes().len() <= max_size(Value)
     pub fn insert(&mut self, key: K, value: V) -> Option<V> {
-        let key_bytes = key.to_bytes();
-        let value_bytes = value.to_bytes();
-
-        match self.version {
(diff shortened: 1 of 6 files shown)


src/btreemap.rs
```diff
@@ -2582,56 +2555,107 @@ mod test {
         });
     }
 
+    // A buggy implementation of storable where the max size is smaller than the serialized size.
```
dsarlis, line 2558:
```suggestion
    // A buggy implementation of storable where the max_size is smaller than the serialized size.
```

dsarlis (approved):

dsarlis:
Oh wait, why did it automatically merge to the base one? Did this change recently?

ielashi:
I think I just did a mistake juggling the different branches. I'll fix this.

ielashi:
I opened #136 to replace this PR.

PR #135, opened 2023-09-20
feat: make BTreeMap V2 the default and automatically migrate V1 maps.

ielashi:
Prior to this commit, initializing a BTreeMap initialized a V1 BTreeMap by default, and only by using `new_v2` and `init_v2` can developers explicitly opt-in to using V2 BTreeMaps.

This commit updates the interface such that a BTreeMap is initialized as a V2 BTreeMap by default, and v1 BTreeMaps are automatically migrated to V2. Developers can still opt-in to using V1 by using `new_v1` and `init_v1`.


src/btreemap.rs
```diff
@@ -2573,77 +2624,20 @@ mod test {
         };
     }
 
-    #[test]
-    #[should_panic(expected = "max_key_size must be <= 0")]
-    fn panics_if_max_key_grows() {
```
ielashi, line 2578: This test is removed as it's already covered by v1_rejects_increases_in_max_key_size

src/btreemap.rs
```diff
@@ -2573,77 +2624,20 @@ mod test {
...
     #[should_panic(expected = "expected an element with length <= 1 bytes, but found 4")]
     fn v2_panics_if_key_is_bigger_than_max_size() {
-        let mut btree = BTreeMap::init_v2(make_memory());
+        let mut btree = BTreeMap::init(make_memory());
         btree.insert(BuggyStruct, ());
     }
 
-    #[test]
-    #[should_panic(expected = "max_value_size must be <= 0")]
-    fn v1_panics_if_value_is_too_large() {
```
ielashi, line 2620: This test is removed as it's already covered by `v1_rejects_larger_value_sizes`

src/btreemap.rs
```diff
@@ -116,25 +116,25 @@ where
...
     }
 
-    /// Initializes a v2 `BTreeMap`.
+    /// Initializes a v1 `BTreeMap`.
     ///
-    /// This is currently exposed only for beta-testing purposes and will be removed in favor of
-    /// using BTreeMap::init directly once V2 is tested well enough.
-    pub fn init_v2(memory: M) -> Self {
+    /// This is primarily exposed for testing and benchmarking.
+    pub fn init_v1(memory: M) -> Self {
```
dsarlis, line 122: nit: Could we get aways with just `pub(crate)` instead?

src/btreemap.rs
```diff
@@ -217,8 +191,39 @@ where
...
+        btree
+    }
+
     /// Loads the map from memory.
     pub fn load(memory: M) -> Self {
+        Self::load_helper(memory, true)
+    }
+
+    // Loads the map from memory, potentially migrating the map from V1 to V2.
+    fn load_helper(memory: M, migrate_to_v2: bool) -> Self {
```
dsarlis, line 226: Question: Would it make sense to have the migration logic guarded by a feature flag only enabled in production builds? The main advantage I see of that approach is that we can drop the boolean arg and have a single load() function.

dsarlis (approved):

src/btreemap.rs
ielashi: Unfortunately because because benchmarking runs the canister's wasm and has no visibility into non-public methods. However, I addressed the issue by updating the method to be only available in testing and if the benchmarking feature is enabled.

src/btreemap.rs
ielashi: It's actually useful to have it as a boolean, because test-cases then have the flexibility to choose which behavior they want (whether to upgrade or not).

PR #142, opened 2023-09-21
feat - Enhance Bound Enum with Utility Methods and Refactor Storable Trait

b3hr4d:
### Description:

This PR introduces utility methods to the `Bound` enum, enhancing its functionality and usability. These methods are used to refactor the `Storable` trait, making it cleaner and more maintainable. 

#### Use-Case:

I am developing a library aimed at simplifying the use of stable memory structures. For example, consider a `TaskTimerEntry` struct defined as:

```rust
pub struct TaskTimerEntry<T> {
    pub time: NanoTimeStamp,
    pub task: T,
}
```

To make this struct bounded, I can now easily specify:

```rust
const BOUND: Bound = Bound::Bounded {
    is_fixed_size: false,
    max_size: 8 + T::BOUND.max_size(),
};
```

#### Key Changes:

1. Introduced `is_bounded`, `max_size`, `is_fixed_size`, `is_within_bounds`, and `verify_fixed_size` methods to `Bound` enum.
2. Refactored the `Storable` trait's `to_bytes_checked` function to utilize the new utility methods, making the code cleaner and more maintainable.

#### Impact on `Storable` Trait:

The utility methods have been integrated into the `Storable` trait, enhancing its readability and maintainability without altering its core functionality.

#### Types of Changes:
- [x] New feature (non-breaking change which adds functionality)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

### DIFF
--- a/src/storable.rs
+++ b/src/storable.rs
@@ -23,22 +23,20 @@ pub trait Storable {
     /// are within the element's bounds.
     fn to_bytes_checked(&self) -> Cow<[u8]> {
         let bytes = self.to_bytes();
-        if let Bound::Bounded {
-            max_size,
-            is_fixed_size,
-        } = Self::BOUND
-        {
-            if is_fixed_size {
-                assert_eq!(
-                    bytes.len(),
-                    max_size as usize,
+        let bound = Self::BOUND;
+
+        if bound.is_bounded() {
+            let max_size = bound.max_size();
+            let is_fixed_size = bound.is_fixed_size();
+
+            if is_fixed_size && !bound.verify_fixed_size(bytes.len()) {
+                panic!(
                     "expected a fixed-size element with length {} bytes, but found {} bytes",
                     max_size,
                     bytes.len()
                 );
-            } else {
-                assert!(
-                    bytes.len() <= max_size as usize,
+            } else if !bound.is_within_bounds(bytes.len()) {
+                panic!(
                     "expected an element with length <= {} bytes, but found {} bytes",
                     max_size,
                     bytes.len()
@@ -82,6 +80,35 @@ impl Bound {
             *max_size
         } else {
             panic!("Cannot get max size of unbounded type.");
+        }
+    }
+
+    /// Returns true if the type is fixed in size, false otherwise.
+    /// Panics if unbounded.
+    pub const fn is_fixed_size(&self) -> bool {
+        if let Bound::Bounded { is_fixed_size, .. } = self {
+            *is_fixed_size
+        } else {
+            panic!("Cannot get fixed size of unbounded type.");
+        }
+    }
+
+    /// Checks if the given size is within the bounds defined.
+    pub const fn is_within_bounds(&self, size: usize) -> bool {
+        match self {
+            Self::Unbounded => true,
+            Self::Bounded { max_size, .. } => size <= *max_size as usize,
+        }
+    }
+
+    /// Checks if the given size matches the fixed size defined.
+    pub const fn verify_fixed_size(&self, size: usize) -> bool {
+        match self {
+            Self::Unbounded => false,
+            Self::Bounded {
+                max_size,
+                is_fixed_size,
+            } => *is_fixed_size && size == *max_size as usize,
         }
     }
 }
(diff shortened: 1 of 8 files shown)


src/storable.rs
```diff
@@ -23,22 +23,20 @@ pub trait Storable {
...
-        if let Bound::Bounded {
-            max_size,
-            is_fixed_size,
-        } = Self::BOUND
-        {
-            if is_fixed_size {
-                assert_eq!(
-                    bytes.len(),
-                    max_size as usize,
+        let bound = Self::BOUND;
```
ielashi, line 26: nit: this is completely subjective, but I actually find the previous code more readable, as it was a simple match statement, so I'd prefer we keep this method as-is.

src/storable.rs
```diff
@@ -70,6 +68,51 @@ pub enum Bound {
     },
 }
 
+impl Bound {
```
ielashi, line 71: Exposing helper methods here is a great idea. I actually exposed a `max_size` function in this file separately as it was useful for developers, but having it be part of the `Bound` struct makes a lot of sense to me.

Question, since I'd rather we keep the public interface as small as possible, do you have use-cases in mind for these methods? I can see the use-case for `max_size` and `is_fixed_size`, but the use-cases for the other methods aren't clear to me.

src/storable.rs
```diff
@@ -70,6 +68,51 @@ pub enum Bound {
...
+        }
+    }
+
+    /// Returns true if the type is fixed in size, false otherwise.
+    /// Panics if unbounded.
+    pub const fn is_fixed_size(&self) -> bool {
+        if let Bound::Bounded { is_fixed_size, .. } = self {
+            *is_fixed_size
+        } else {
+            panic!("Cannot get fixed size of unbounded type.");
```
ielashi, line 92: I'd return false here rather than panic.

src/storable.rs
b3hr4d: Thank you for your feedback!

I completely understand the desire to keep the public interface as minimal as possible. The primary motivation behind adding the additional utility methods was to enhance the flexibility and usability of the Bound enum, both within the library and for external use.

In my project, I mainly utilize the max_size and is_fixed_size methods, as you've rightly pointed out. These methods are crucial for managing memory bounds in my TaskTimerEntry struct and similar data structures.

The other methods (is_within_bounds and verify_fixed_size) were added to provide additional utility that might be beneficial in more complex use-cases, although they are not directly used in my project. Given your preference for a minimal public interface, I agree that focusing on max_size and is_fixed_size would be the most practical approach.

Would you like me to update the pull request to include only these two methods, while removing the others?

src/storable.rs
b3hr4d: Thank you for the suggestion. I agree that if we decide to remove the other utility methods and keep the interface minimal, returning `false` instead of panicking would make more sense. This approach would be more in line with the principle of least astonishment for the users of the library.

I'll go ahead and update the `is_fixed_size` method to return `false` for unbounded types.

src/storable.rs
ielashi: Thanks for your understanding @b3hr4d. Yes, I'd suggest we keep only the ones we have concrete use-cases for (`max_size` and `is_fixed_size`). We can easily add additional helper methods as needed in the future.

src/storable.rs
```diff
@@ -70,6 +70,27 @@ pub enum Bound {
...
+    pub const fn max_size(&self) -> u32 {
+        if let Bound::Bounded { max_size, .. } = self {
+            *max_size
+        } else {
+            panic!("Cannot get max size of unbounded type.");
+        }
+    }
+
+    /// Returns true if the type is fixed in size, false otherwise.
+    /// Panics if unbounded.
```
ielashi, line 84: I think this line is no longer true.
```suggestion
```

ielashi (approved):
LGTM, thanks! It'd be great if you could update the description to fit the new changes.

PR #143, opened 2023-09-22
feat: comprehensive fuzzing for BTreeMap

ielashi:
Adds a fuzz test that can run indefinitely comparing the results of a stable BTreeMap to a standard BTreeMap.

I ended up using `proptest` rather than the conventional `cargo fuzz` for two reasons:

1. `cargo fuzz` requires nightly rust.
2. With proptest, we can have a short version of the fuzz test that runs as part of CI (which is included in this commit).

### DIFF
--- a/src/btreemap/proptests.rs
+++ b/src/btreemap/proptests.rs
@@ -134,6 +134,8 @@ fn map_upper_bound_iter(#[strategy(pvec(0u64..u64::MAX -1 , 10..100))] keys: Vec
     });
 }
 
+// Given an operation, executes it on the given stable btreemap and standard btreemap, verifying
+// that the result of the operation is equal in both btrees.
 fn execute_operation<M: Memory>(
     std_btree: &mut StdBTreeMap<Vec<u8>, Vec<u8>>,
     btree: &mut BTreeMap<Vec<u8>, Vec<u8>, M>,
--- a/src/btreemap/proptests.rs
+++ b/src/btreemap/proptests.rs
@@ -23,7 +23,7 @@ enum Operation {
 // A custom strategy that gives unequal weights to the different operations.
 // Note that `Insert` has a higher weight than `Remove` so that, on average, BTreeMaps
 // are growing in size the more operations are executed.
-fn op_strategy() -> impl Strategy<Value = Operation> {
+fn operation_strategy() -> impl Strategy<Value = Operation> {
     prop_oneof![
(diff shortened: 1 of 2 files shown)


src/btreemap/proptests.rs
```diff
@@ -1,17 +1,89 @@
...
+#[proptest(cases = 10)]
+fn comprehensive(#[strategy(pvec(op_strategy(), 100..5_000))] ops: Vec<Operation>) {
+    let mem = make_memory();
+    let mut btree = BTreeMap::new(mem);
+    let mut std_btree = StdBTreeMap::new();
+
+    // Execute all the operations, validating that the stable btreemap behaves similarly to a std
+    // btreemap.
+    for op in ops.into_iter() {
+        execute_operation(&mut std_btree, &mut btree, op);
```
AlexandraZapuc, line 52: nit: I would rename this function. Maybe `validate_trees_after_operation`? Or something else that makes it obvious the purpose of the function is to verify the trees.

src/btreemap/proptests.rs
```diff
@@ -1,17 +1,89 @@
...
+    Insert { key: Vec<u8>, value: Vec<u8> },
+    Iter { from: usize, len: usize },
+    Get(usize),
+    Remove(usize),
+}
+
+// A custom strategy that gives unequal weights to the different operations.
+// Note that `Insert` has a higher weight than `Remove` so that, on average, BTreeMaps
+// are growing in size the more operations are executed.
+fn op_strategy() -> impl Strategy<Value = Operation> {
```
AlexandraZapuc, line 26: nit: rename `operation_strategy`

src/btreemap/proptests.rs
```diff
@@ -1,17 +1,89 @@
...
+    let mut runner = TestRunner::default();
+
+    let mem = make_memory();
+    let mut btree = BTreeMap::new(mem);
+    let mut std_btree = StdBTreeMap::new();
+
+    let mut i = 0;
+
+    loop {
+        let op = op_strategy().new_tree(&mut runner).unwrap().current();
```
AlexandraZapuc, line 77: nit: Rename `op` to `operation`.

AlexandraZapuc (approved):

src/btreemap/proptests.rs
ielashi: Actually, the purpose of this method _is_ to execute an operation, so the current naming is accurate.

PR #145, opened 2023-09-26
feat: impl Storable for MemoryId

witter-deland:


src/memory_manager.rs
```diff
@@ -518,6 +518,21 @@ impl MemoryId {
...
+        self.0.to_bytes()
+    }
+
+    fn from_bytes(bytes: std::borrow::Cow<[u8]>) -> Self {
+        Self(u8::from_bytes(bytes))
+    }
+
+    const BOUND: crate::storable::Bound = crate::storable::Bound::Bounded {
+        max_size: 1,
+        is_fixed_size: true,
```
roman-kashitsyn, line 532: I have the same concerns as I had here: https://github.com/dfinity/stable-structures/pull/99#pullrequestreview-1516297448

With this change, we commit to MemoryId being a u8 forever.
If there is a good use-case for storing MemoryId in a stable structure, I'd prefer to represent it as [LEB128](https://en.wikipedia.org/wiki/LEB128) encoded variable-size integer.

witter-deland:
I'll use u8 instead :)

PR #148, opened 2023-09-28
fix: support for migrating from bounded type to unbounded type.

ielashi:
We'd like to support developers changing the bounds of their types from bounded to unbounded. There were some changes that needed to be made in this commit in order to make this migration work.

### DIFF
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -2803,7 +2803,7 @@ mod test {
         let btree: BTreeMap<T2, T2, _> = BTreeMap::init(btree.into_memory());
         assert_eq!(btree.get(&T2), Some(T2));
 
-        // Relaod the BTree again with bounded type.
+        // Reload the BTree again with bounded type.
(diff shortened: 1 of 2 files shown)


src/btreemap.rs
```diff
@@ -2712,4 +2752,59 @@ mod test {
...
+
+        // Migrate to v2 and the unbounded type.
+        let btree: BTreeMap<T2, T2, _> = BTreeMap::init(btree.into_memory());
+        btree.save();
+
+        // Reload the BTree again and try to read the value.
+        let btree: BTreeMap<T2, T2, _> = BTreeMap::init(btree.into_memory());
+        assert_eq!(btree.get(&T2), Some(T2));
+
+        // Relaod the BTree again with bounded type.
```
dsarlis, line 2806:
```suggestion
        // Reload the BTree again with bounded type.
```

dsarlis (approved):

PR #160, opened 2023-12-11
feat: Add benchmark to StableStructures CI

dragoljub-duric:


benchmark-canisters/src/memory_manager.rs
```diff
@@ -56,3 +57,21 @@ pub fn memory_manager_overhead() -> BenchResult {
         }
     })
 }
+
+/// Benchmarks the `MemoryManager` by allocating a large number of buckets.
+#[ic_cdk_macros::query]
+pub fn memory_manager_buckets_allocation() -> BenchResult {
```
ielashi, line 63: A couple of points:

1. I don't understand what the intent is of this benchmark. What more does it add compares to the ones that we already have?
2. This is totally unrelated to adding benchmarks to CI, so let's please move it out of this PR.

.github/workflows/ci.yml
```diff
@@ -70,3 +70,26 @@ jobs:
       - name: Run test
         run: |
           bash examples/test.sh
+
+  benchmark:
```
ielashi, line 74: Now that we're adding new jobs to CI, maybe it's time to add something like the `checks-pass` in the bitcoin canister repo? We can then configure this repo to ensure that `checks-pass` passes before PRs are merged.

benchmark-canisters/src/memory_manager.rs
ielashi: If the goal is to understand the performance regression with the memory.free changes, then we need a scenario where we declare tens of thousands of buckets, and then we benchmark how long it takes to call `memory.grow`. But, again, that's a different PR.

.github/workflows/ci.yml
dragoljub-duric: Done.

ielashi (approved):
LGTM, thank you!

PR #162, opened 2023-12-13
feat: impl Storable for candid:Principal

sesi200:
It would be very convenient to have `Storable` support for `candid::Principal`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "anstream"
-version = "0.5.0"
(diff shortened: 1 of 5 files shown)


ielashi (changes requested):
We have discussed this before, and we cannot depend on candid in stable-structures. To fix this properly, we proposed moving the `Principal` type into a separate crate `ic-principal`, that contains only the `Principal`, and then we can depend on it.

`candid` is a fat dependency and different versions of it aren't compatible with each other, so depending on it would cause more problems than it solves.

PR #163, opened 2023-12-13
feat: impl Storable for ic_principal::Principal

sesi200:
It would be very convenient to have `Storable` support for `ic_principal::Principal`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -440,6 +440,7 @@ dependencies = [
  "hex",
  "ic-cdk",
  "ic-cdk-macros",
+ "ic_principal",
  "lazy_static",
  "maplit",
  "profiler",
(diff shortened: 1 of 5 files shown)


ielashi (commented):
I was expecting `ic_principal` to be a very light-weight crate that has no dependencies, but looking at the crate that's not the case. `ic_principal` has the following dependencies:

```
[dependencies]
crc32fast = "1.3.0"
data-encoding = "2.3.2"
hex = "0.4.3"
sha2 = "0.10.1"
thiserror = "1.0.30"
```

We've been very careful to not introduce dependencies in stable-structures unless there's a strong reason to. Depending on `ic-principal` makes sense, but also pulling in all these other dependencies makes less sense to me. In principle, none of these dependencies are necessary for this crate. Maybe these dependencies can be moved into features so that not all consumers have to pull those dependencies too?

ielashi (approved):
LGTM, thanks!

PR #164, opened 2023-12-13
perf: Add benchmark test with allocation of many buckets to StableStructures

dragoljub-duric:


benchmark-canisters/src/memory_manager.rs
```diff
@@ -56,3 +56,19 @@ pub fn memory_manager_overhead() -> BenchResult {
         }
     })
 }
+
+/// Benchmarks the `MemoryManager` by allocating a large number of buckets.
```
ielashi, line 60:
```suggestion
/// Benchmarks the `MemoryManager`'s `grow` method.
```

ielashi (approved):

PR #174, opened 2024-01-17
refactor: add config file for benchmarks

ielashi:
This PR is another step towards transforming our benchmarking logic into a general-purpose tool that others can use.

Rather than hard-coding everything in `benchmark.rs`, the configuration is moved into a config file called `bench.yml`.

The eventual goal is to evolve the current benchmarking work into a tool that works like this:

1. User defines a configuration file (e.g. `bench.yml`)
2. User installs our benchmarking tool (e.g. `cargo install canbench`)
3. User can now run benchmarks using cargo (e.g. `cargo canbench`)

Note that users won't need to write any code and the file `benchmark.rs` will soon be completely removed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -692,6 +692,7 @@ dependencies = [
  "maplit",
  "profiler",
  "proptest",
+ "serde_yaml",
  "tempfile",
  "test-strategy",
 ]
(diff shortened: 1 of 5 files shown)


AlexandraZapuc (approved):
Lgtm!

PR #183, opened 2024-01-26
chore: bump `candid` and `ic-cdk` versions

hpeebles:


ielashi:
@hpeebles can you provide context on why we need this PR, and why are we bumping the version? stable-structures doesn't depend on neither candid nor ic-cdk.

hpeebles:
Ah I didn't realise they were only dev dependencies.
I'm in the process of trying to bump candid and ic-cdk within OC and the build is failing because some of our dependencies are not on the latest versions so I've just been going through them and bumping them.
Given that they are only dev dependencies here I'm happy to close this PR.

PR #188, opened 2024-02-01
feat: Implement into_memory for memory manager

bitdivine:
# Motivation
When a memory manager is created, the constructor (`init()`) consumes the undertlying memory and there is no way of getting it back.

A workaround is to clone a reference to the memory and keep the reference around, however that opens the possibility that the underlying memory is changed without going via the memory manager.  It would be better if the memory manager could be made to surrender the underlying memory.

# Changes
- Add `into_inner()` to `MemoryManager` (and to the inner data structure).

# Tests
- A unit test is included to demonstrate that memory can be returned.


ielashi:
@bitdivine Can you share what your use-case is?

bitdivine:
@ielashi Sure thing:  I am migrating from raw memory to stable structures and back again.  Migrating back involves getting the raw memory back from the managed memory - or keeping a second reference.  With a wrapper that keeps both managed memory and a second pointer, I also found that this simplifies unit testing schema upgrades.  Restarting the "canister" in tests is then just a matter of getting the raw memory back, and that raw memory is then loaded in the upgraded canister.  I don't need an entire replica for this.  This was not the original motivation but was a nice bonus! :-)

src/tests.rs
```diff
@@ -48,3 +49,11 @@ fn vec_exposes_init_err() {
     #[allow(unused_imports)]
     use crate::vec::InitError;
 }
+
+#[test]
+fn should_be_able_to_recover_memory_from_memory_manager() {
```
ielashi, line 54: Can we also add a test where there's an underlying reference and `into_memory` returns `None`?

bitdivine: Certainly.  Done.

bitdivine:
It looks as if CI failed because it could not find a comment in the benchmark output.  The comment logic in CI was not present in main when I made the fork so I have pulled in the latest changes from main. Hopefully that will fix it.

ielashi (approved):

ielashi:
> It looks as if CI failed because it could not find a comment in the benchmark output. The comment logic in CI was not present in main when I made the fork so I have pulled in the latest changes from main. Hopefully that will fix it.
> 
> No such luck. Here is the error: Screenshot from 2024-02-20 13-31-02

@bitdivine The issue was actually more complex to fix than excepted, but it's finally resolved. PR LGTM! :)

bitdivine:
@ielashi Thank you!  I do not have permission to merge but CI is green.

PR #190, opened 2024-02-07
perf: BTreeMap: make `range().count()` more efficient

dragoljub-duric:
Problem:
BTreeMap::Iterator::count() implicitly uses BTreeMap::Iterator::next() which in some cases loads the map entry's value non-lazy. 

Solution:
Explicitly reimplement BTreeMap::Iterator::count() so it does not load the value.


src/btreemap/iter.rs
```diff
@@ -71,19 +73,10 @@ where
...
-    K: Storable + Ord + Clone,
-    V: Storable,
-    M: Memory,
-{
-    type Item = (K, V);
-
-    fn next(&mut self) -> Option<Self::Item> {
-        match self.cursors.pop() {
-            Some(Cursor::Address(address)) => {
+    fn update_cursors_for_iteration(&mut self, cursor: Cursor<K>) {
```
dragoljub-duric, line 77: This part is extracted from previous implementation of `fn next()` so it can be reused for both `next()` and `next_without_loading_value()`.

dragoljub-duric:
@ielashi I can add a benchmark to compare it with the previous implementation, but I am not sure that is important. WDYT?

ielashi:
> @ielashi I can add a benchmark to compare it with the previous implementation, but I am not sure that is important. WDYT?

It is absolutely important :) How else will we know that our change has improved anything?

dragoljub-duric:
@ielashi The bench is added:)

src/btreemap/iter.rs
```diff
@@ -140,21 +128,82 @@ where
...
+
+impl<K, V, M> Iterator for Iter<'_, K, V, M>
+where
+    K: Storable + Ord + Clone,
+    V: Storable,
+    M: Memory,
+{
+    type Item = (K, V);
+
+    fn next(&mut self) -> Option<Self::Item> {
```
ielashi, line 170: This function and `next_without_loading_value` look awfully similar. I wonder if we can somehow combine the two to avoid duplicate logic ;)

src/btreemap.rs
```diff
@@ -2317,6 +2317,21 @@ mod test {
         });
     }
 
+    #[test]
+    fn iter_count_test() {
```
ielashi, line 2321: proptests are almost always preferable to standard unit tests :)

benchmarks/src/btreemap.rs
```diff
@@ -219,6 +220,34 @@ pub fn btreemap_insert_10mib_values() -> BenchResult {
     })
 }
 
+#[bench]
+pub fn btreemap_iter_count() -> BenchResult {
```
ielashi, line 224: I'd also be curious how performance is like when the value is very small (e.g. a `u8`). Also, ideally, the benchmarks are added in one PR, and in a follow-up PR we make the improvements to the benchmarks.

src/btreemap/iter.rs
```diff
@@ -140,21 +128,82 @@ where
...
 
-                Some((key, V::from_bytes(Cow::Owned(encoded_value))))
+                return Some(());
             }
-            None => {
-                // The cursors are empty. Iteration is complete.
-                None
+        }
+
+        self.next().map(|_| ())
```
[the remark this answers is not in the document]
dragoljub-duric: Thank you for noticing, that was a mistake.

benchmarks/src/btreemap.rs
dragoljub-duric: The benchmarks are now in separate [PR](https://github.com/dfinity/stable-structures/pull/191), and new one is added.

src/btreemap.rs
dragoljub-duric: Done.

src/btreemap/iter.rs
dragoljub-duric: Done.

dragoljub-duric: What do you think about adding a new public fn to Iter `next_key()` which will be almost the same as `next_without_loading_value`? BTreeMap is an ordered set, and providing such functionality without loading values may be useful.

src/btreemap.rs
```diff
@@ -3029,4 +3030,25 @@ mod test {
         let btree: BTreeMap<T, T, _> = BTreeMap::init(btree.into_memory());
         assert_eq!(btree.get(&T), Some(T));
     }
+
+    proptest! {
```
ielashi, line 3034: Let's please move the `proptest` to `btreemap/proptests.rs`, and also let's use the `#[proptest]` macro, as it allows rust to format the code.

src/btreemap/iter.rs
```diff
@@ -6,12 +6,14 @@ use crate::{types::NULL, Address, Memory, Storable};
 use std::borrow::Cow;
 use std::ops::{Bound, RangeBounds};
 
+#[derive(Clone)]
```
ielashi, line 9: Cloning is something we explicitly do not want to support. Cloning nodes/cursors is a heavy operation and we should avoid it.

src/btreemap/iter.rs
ielashi: `next_key` is not a native function available in `Iterator`, so it cannot be added here IIUC.

src/btreemap/iter.rs
```diff
@@ -140,21 +128,67 @@ where
...
+    K: Storable + Ord + Clone,
+    V: Storable,
+    M: Memory,
+{
+    type Item = (K, V);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // It is safe to unwrap value because when the key exists the
+        // value will exist as well, and it will be loaded.
+        self.next_internal(true).map(|(k, v)| (k, v.unwrap()))
```
ielashi, line 180: This is a step in the right direction, but I'm not a big fan of having unwraps :) There's a way to avoid unwraps as well as cloning. I'll let you think about it - ping me if you get stuck.

src/btreemap/node.rs
```diff
@@ -46,7 +46,7 @@ pub type Entry<K> = (K, Vec<u8>);
 /// 2. `V2`, which supports both bounded and unbounded types.
 ///
 /// See `v1.rs` and `v2.rs` for more details.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
ielashi, line 49: ditto: no cloning please

src/btreemap/node.rs
```diff
@@ -434,7 +434,7 @@ impl NodeHeader {
 }
 
 // The value in a K/V pair.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
ielashi, line 437: ditto

src/btreemap/iter.rs
dragoljub-duric: My idea was that `next_key()` become the function of `Iter` not `Iterator`.

src/btreemap/iter.rs
dragoljub-duric: Cloning is removed, but for the unwraps I am not sure in which way you want me to continue, can you please tell me?

src/btreemap/iter.rs
ielashi: Unless there's a strong need for it, I wouldn't worry about it. There are always new things that can be added - better to be as lazy as possible when it comes to these things :)

src/btreemap/iter.rs
ielashi: Sure, here's what I'd do:

Instead of `next_internal`, I'd define a method called `next_map`. This is a method that iterates through the tree, and the caller can provide a `map` closure that takes a node and entry point idx and maps it to whatever it wants to return. Here's the signature:

```
fn next_map<T, F: Fn(&Node<K>, usize) -> T>(&mut self, map: F) -> Option<T> {
```

Then, `next` can be implemented like this:

```
impl<K, V, M> Iterator for Iter<'_, K, V, M>
where
    K: Storable + Ord + Clone,
    V: Storable,
    M: Memory,
{
    type Item = (K, V);

    fn next(&mut self) -> Option<Self::Item> {
        self.next_map(|node, entry_idx| {
            let (key, encoded_value) = node.entry(entry_idx, self.map.memory());
            (key, V::from_bytes(Cow::Owned(encoded_value)))
        })
    }
}
```

The count can be implemented similarly, but the map function will simply do nothing.

src/btreemap/proptests.rs
```diff
@@ -144,6 +144,22 @@ fn map_upper_bound_iter(#[strategy(pvec(0u64..u64::MAX -1 , 10..100))] keys: Vec
     });
 }
 
+#[proptest]
+fn iter_count_test(#[strategy(0..250u8)] start: u8, #[strategy(1..250u8)] size: u8) {
+    let end = std::cmp::min(start as u64 + size as u64, 255) as u8;
```
ielashi, line 149: How long does this take to run locally for you? I just ran it and it took for me 193 seconds. We shouldn't write tests that take this much time unless absolutely necessary.

You can reduce the speed of this test by simply running it for less cases. For example, you can annotate it with something like this:

```
#[proptest(cases = 10)]
```

Also, I believe this is closer to what you actually want:

```
fn iter_count_test(#[strategy(0..250u8)] start: u8, #[strategy(#start..255u8)] end: u8) {
```

ielashi (approved):

PR #191, opened 2024-02-08
perf: Add benchmarks for BtreeMap::Iterator::count()

dragoljub-duric:


benchmarks/src/btreemap.rs
```diff
@@ -219,6 +221,53 @@ pub fn btreemap_insert_10mib_values() -> BenchResult {
     })
 }
 
+#[bench]
+pub fn btreemap_iter_count() -> BenchResult {
```
ielashi, line 225:
```suggestion
pub fn btreemap_iter_count_small_values() -> BenchResult {
```

benchmarks/src/btreemap.rs
```diff
@@ -219,6 +221,53 @@ pub fn btreemap_insert_10mib_values() -> BenchResult {
...
+    }
+
+    benchmark(|| {
+        for i in 0..size {
+            for j in i + 1..size {
+                btree
+                    .range((Bound::Included(i), Bound::Included(j)))
+                    .count();
+            }
+        }
```
ielashi, line 239: What do we gain by having this loop. Why not simplify to something like this?

```suggestion
        btree.range((Bound::Included(0), Bound::Included(size))).count();
```

benchmarks/src/btreemap.rs
```diff
@@ -219,6 +221,53 @@ pub fn btreemap_insert_10mib_values() -> BenchResult {
...
+#[bench]
+pub fn btreemap_iter_count_10mib_values() -> BenchResult {
+    let mut btree = BTreeMap::new(DefaultMemoryImpl::default());
+
+    // Insert 200 10MiB values.
+    let mut rng = Rng::from_seed(0);
+    let mut values = vec![];
+    for _ in 0..200 {
+        values.push(rng.iter(Rand::rand_u8).take(10 * 1024).collect::<Vec<_>>());
+    }
```
ielashi, line 252: You can also just push to the btree directly in the same loop:

```suggestion
    for i in 0..200 {
        btree.insert(i, rng.iter(Rand::rand_u8).take(10 * 1024).collect::<Vec<_>>()));
    }
```

benchmarks/src/btreemap.rs
```diff
@@ -219,6 +221,53 @@ pub fn btreemap_insert_10mib_values() -> BenchResult {
...
+
+#[bench]
+pub fn btreemap_iter_count_10mib_values() -> BenchResult {
+    let mut btree = BTreeMap::new(DefaultMemoryImpl::default());
+
+    // Insert 200 10MiB values.
+    let mut rng = Rng::from_seed(0);
+    let mut values = vec![];
+    for _ in 0..200 {
+        values.push(rng.iter(Rand::rand_u8).take(10 * 1024).collect::<Vec<_>>());
```
ielashi, line 251: I also don't think we really need to generate random bytes. Just a large vector of zeros would be sufficient for what we want to benchmark here.

benchmarks/src/btreemap.rs
dragoljub-duric: We are checking if the count works for all possible ranges in the BTreeMap. We are not getting a lot, but we are a bit more sure it works properly. And since there are not that many ranges I was thinking that it is good to have it this way.

dragoljub-duric:
@ielashi PTAL

benchmarks/src/btreemap.rs
```diff
@@ -219,6 +221,39 @@ pub fn btreemap_insert_10mib_values() -> BenchResult {
     })
 }
 
+#[bench]
+pub fn btreemap_iter_count_small_values() -> BenchResult {
+    let mut btree = BTreeMap::new(DefaultMemoryImpl::default());
+    let size: u32 = 2000;
```
ielashi, line 227: I'd bump this to 10,000. We can be liberal in benchmarks :)

ielashi (approved):

PR #192, opened 2024-02-08
fix: failing edge cases in `Writer` tests

dragoljub-duric:
Addresses edge cases in two of our `Writer` proptests:

1. `should_only_request_min_number_of_pages_required`

    The test fails when `bytes` is an empty vector:

```
thread 'writer::tests::should_only_request_min_number_of_pages_required' panicked at src/writer/tests.rs:66:9: 
 assertion left == right failed 
 left: 0 
 right: 1
minimal failing input: buffer_size = None, bytes = [], offset = 1
```

2. `should_write_single_slice`

    The test also fails when `bytes` is an empty vector:

```
thread 'writer::tests::should_write_single_slice' panicked at src/writer/tests.rs:6:1:
Test failed: read: out of bounds.
minimal failing input: buffer_size = None, bytes = [], offset = 1
        successes: 0
        local rejects: 0
        global rejects: 0
```


src/writer/tests.rs
```diff
@@ -61,7 +61,11 @@ proptest! {
         }
 
         let capacity_pages = memory.size();
-        let min_pages_required = (offset + bytes.len() as u64 + WASM_PAGE_SIZE - 1) / WASM_PAGE_SIZE;
+        let min_pages_required = if bytes.is_empty() {
+            0
+        }else{
+            (offset + bytes.len() as u64 + WASM_PAGE_SIZE - 1) / WASM_PAGE_SIZE
+        };
```
ielashi, line 68: nit: I think we can use div_ceil here for clarity.

```suggestion
        } else {
            u64::div_ceil(offset + bytes.len() as u64, WASM_PAGE_SIZE)
        };
```

ielashi (approved):

ielashi:
@dragoljub-duric  Also, proptests should include a file in `proptest-regressions`. Please include it in this PR so that proptests remembers to test this specific case in the future.

src/writer/tests.rs
```diff
@@ -18,7 +18,9 @@ proptest! {
             writer.flush().unwrap();
         }
         let mut buf = vec![0; bytes.len()];
-        memory.read(offset, &mut buf);
+        if !bytes.is_empty(){
```
dragoljub-duric, line 21: @ielashi After adding the file to proptest-regressions it revealed that `should_write_single_slice` is flaky for the same input. Hence to pass CI both tests should be fixed in a single PR.

dragoljub-duric:
@ielashi PTAL.

src/writer/tests.rs
ielashi: @dragoljub-duric Thanks for noticing. Looking.

ielashi: Rather than adding this exception, I'd suggest we change the proptest to not include bytes of zero length.

nit: let's also update the PR description/title accordingly.

ielashi (changes requested):

ielashi (approved):
LGTM. Per our discussion in another PR around clearer communication, I made some adjustments to the PR title/description.

1. PR title was too long. I moved some of the details into the description
2. The tests are not actually flaky, so I removed that term from the title. Flaky means that the same test with the same input sometimes succeeds and sometimes fails. In this case, there's an edge case that has been found by proptests, but for that edge case it fails deterministically.

PR #193, opened 2024-02-09
chore: Upgrade Rust version

dragoljub-duric:
Problem:
CI was failing because rust >= 1.74 was required.

Solution:
Upgrade the Rust version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rust: [ 1.73.0 ]
+        rust: [ 1.76.0 ]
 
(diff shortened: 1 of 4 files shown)


ielashi:
@dragoljub-duric Clear communication is a very valuable skill to develop as an engineer - PR titles and descriptions are especially important components of that.

Reading the current PR title/description, I see:

* Upgrade Rust in CI - this is misleading, since you're upgrading Rust everywhere and not only in CI.
* There's no description, so I'm left wondering why you're making this upgrade. Is it just a casual upgrade, was there something blocking you that needed such an upgrade, some other reason?

I'll approve but I'd encourage you to enhance the title/description and take that into account in future PRs.

ielashi (approved):

PR #195, opened 2024-02-14
feat: Implement Storable for tuples with three elements

dragoljub-duric:
Currently, Storable is implemented for two-element tuples (A, B), but not for three-element tuples (A, B, C). Adding support for three-element tuples would help with usability.

### DIFF
--- a/src/storable.rs
+++ b/src/storable.rs
@@ -502,6 +502,7 @@ where
         match Self::BOUND {
             Bound::Bounded { max_size, .. } => {
                 let mut bytes = vec![0; max_size as usize];
+
                 let a_bytes = self.0.to_bytes();
                 let a_bounds = bounds::<A>();
                 let a_max_size = a_bounds.max_size as usize;
@@ -515,10 +516,10 @@ where
                     &a_bounds,
                 );
 
-                let b_c_bytes = <(B, C)>::to_bytes(&(self.1, self.2));
+                let b_c_bytes = (self.1, self.2).to_bytes();
 
                 bytes[a_max_size + a_size_len..a_max_size + a_size_len + b_c_bytes.len()]
-                    .copy_from_slice(b_c_bytes.into_owned().as_slice());
+                    .copy_from_slice(b_c_bytes.borrow());
 
                 Cow::Owned(bytes)
             }
--- a/src/storable.rs
+++ b/src/storable.rs
@@ -497,26 +497,21 @@ where
 // When serialized `entry` is saved in `bytes` on indices `[start, end)`
 // the function will return index `end` - the first index after
 // `start` that is not occupied with the serialization of `entry`.
-fn serialize_bounded_with_size<T>(entry: &T, bytes: &mut [u8], start: usize) -> usize
+fn serialize_bounded_with_size<T>(entry_bytes: &[u8], bytes: &mut [u8], start: usize) -> usize
 where
     T: Storable,
 {
     let bounds = bounds::<T>();
     let size_len = bytes_to_store_size(&bounds) as usize;
     let max_size = bounds.max_size as usize;
-    let entry_bytes = entry.to_bytes();
     debug_assert!(entry_bytes.len() <= max_size);
+    let actual_size = entry_bytes.len();
 
-    encode_size(
-        &mut bytes[start..start + size_len],
-        entry_bytes.len(),
-        &bounds,
-    );
+    encode_size(&mut bytes[start..start + size_len], actual_size, &bounds);
 
-    bytes[start + size_len..start + size_len + entry_bytes.len()]
-        .copy_from_slice(entry_bytes.borrow());
+    bytes[start + size_len..start + size_len + actual_size].copy_from_slice(entry_bytes);
 
-    start + max_size + size_len
+    start + actual_size + size_len
 }
 
 // Deserialize the struct starting at index `start` in `bytes`.
@@ -528,14 +523,26 @@ where
     T: Storable,
 {
     let bounds = bounds::<T>();
-    let max_size = bounds.max_size as usize;
     let size_len = bytes_to_store_size(&bounds) as usize;
-    let len = decode_size(&bytes[start..start + size_len], &bounds);
+    let actual_size = decode_size(&bytes[start..start + size_len], &bounds);
 
     let a = T::from_bytes(Cow::Borrowed(
-        &bytes[start + size_len..start + size_len + len],
+        &bytes[start + size_len..start + size_len + actual_size],
     ));
-    (a, start + max_size + size_len)
+    (a, start + actual_size + size_len)
+}
+
+fn get_size_len<T>() -> usize
+where
+    T: Storable,
+{
+    match T::BOUND {
+        Bound::Bounded { .. } => {
+            let bounds = bounds::<T>();
+            bytes_to_store_size(&bounds) as usize
+        }
+        Bound::Unbounded => todo!("Serializing tuples with unbounded types is not yet supported."),
+    }
 }
 
 impl<A, B, C> Storable for (A, B, C)
@@ -546,12 +553,23 @@ where
 {
     fn to_bytes(&self) -> Cow<[u8]> {
         match Self::BOUND {
-            Bound::Bounded { max_size, .. } => {
-                let mut bytes = vec![0; max_size as usize];
+            Bound::Bounded { .. } => {
+                let a_bytes = self.0.to_bytes();
+                let b_bytes = self.1.to_bytes();
+                let c_bytes = self.2.to_bytes();
 
-                let a_end = serialize_bounded_with_size(self.0.borrow(), &mut bytes, 0);
-                let b_end = serialize_bounded_with_size(self.1.borrow(), &mut bytes, a_end);
-                serialize_bounded_with_size(self.2.borrow(), &mut bytes, b_end);
+                let output_size = a_bytes.len()
+                    + get_size_len::<A>()
+                    + b_bytes.len()
+                    + get_size_len::<B>()
+                    + c_bytes.len()
+                    + get_size_len::<C>();
+
+                let mut bytes = vec![0; output_size];
+
+                let a_end = serialize_bounded_with_size::<A>(a_bytes.borrow(), &mut bytes, 0);
+                let b_end = serialize_bounded_with_size::<B>(b_bytes.borrow(), &mut bytes, a_end);
+                serialize_bounded_with_size::<C>(c_bytes.borrow(), &mut bytes, b_end);
                 Cow::Owned(bytes)
             }
             _ => todo!("Serializing tuples with unbounded types is not yet supported."),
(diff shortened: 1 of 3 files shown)


dragoljub-duric:
@ielashi The idea I have in mind for the unbounded tuple is to reuse functions `serialize_bounded_with_size` and `deserialize_bounded_with_size`, instead of relying on Bounds, we assume that the first 8 bytes represent the length of the entry. WDYT?

dragoljub-duric:
[Here](https://github.com/dfinity/stable-structures/pull/196/files) is the sketch of follow-up for Unbounded types.

src/storable.rs
```diff
@@ -492,6 +492,102 @@ where
...
+    (a, start + max_size + size_len)
+}
+
+impl<A, B, C> Storable for (A, B, C)
+where
+    A: Storable,
+    B: Storable,
+    C: Storable,
+{
+    fn to_bytes(&self) -> Cow<[u8]> {
```
ielashi, line 538: Would it not make sense to re-use code between this implementation and the `(A, B)` implementation?

src/storable.rs
```diff
@@ -492,6 +492,102 @@ where
     };
 }
 
+fn serialize_bounded_with_size<T>(entry: &T, bytes: &mut [u8], start: usize) -> usize
```
ielashi, line 495: Docs please (again, related to the point of communication that we discussed). It isn't clear, just reading this function signature what it's meant to return.

src/storable.rs
```diff
@@ -492,6 +492,102 @@ where
...
+        &bounds,
+    );
+
+    bytes[start + size_len..start + size_len + entry_bytes.len()]
+        .copy_from_slice(entry_bytes.borrow());
+
+    start + max_size + size_len
+}
+
+fn deserialize_bounded_with_size<T>(bytes: &[u8], start: usize) -> (T, usize)
```
ielashi, line 517: ditto

src/storable.rs
```diff
@@ -492,6 +492,102 @@ where
...
+
+    const BOUND: Bound = {
+        match (A::BOUND, B::BOUND, C::BOUND) {
+            (Bound::Bounded { .. }, Bound::Bounded { .. }, Bound::Bounded { .. }) => {
+                let a_bounds = bounds::<A>();
+                let b_bounds = bounds::<B>();
+                let c_bounds = bounds::<C>();
+
+                Bound::Bounded {
+                    max_size: a_bounds.max_size
```
ielashi, line 575: Do we need to actually store the sizes if all the elements are fixed? I think we can get away with not storing them at all in that case to save some bytes.

src/storable.rs
ielashi: I think it's possible, but per my other comment, I think the current implementation of `(A, B)` is inefficient, as we're needlessly storing the sizes when elements are fixed, so better keep this implementation separate so that we can optimize it better.

src/storable.rs
dragoljub-duric: Done.

src/storable.rs
dragoljub-duric: Do you think that we do not store `Bound` for `tuples (A, B, C)` at all?

dragoljub-duric: If that is your idea I agree, I can do that.

ielashi: I'm not sure I understand your question, but here's an example: suppose you're storing `(u8, u8, u8)`. `u8` is a fixed type, in the sense its size is _always_ one byte. In that case, we don't need to store the size in the tuple encoding because we already know it.

dragoljub-duric: Done, PTAL.

src/storable.rs
```diff
@@ -492,6 +492,123 @@ where
...
+impl<A, B, C> Storable for (A, B, C)
+where
+    A: Storable,
+    B: Storable,
+    C: Storable,
+{
+    fn to_bytes(&self) -> Cow<[u8]> {
+        match Self::BOUND {
+            Bound::Bounded { max_size, .. } => {
+                let mut bytes = vec![0; max_size as usize];
```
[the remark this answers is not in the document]
dragoljub-duric: TBH, my understanding about Bounded Types was that if we load them from Stable Memory and change them, we can store them in the same place we loaded them. With this change I am not sure why we will need to distinguish Bounded and Unbounded types for tuples at all.

src/storable.rs
```diff
@@ -492,6 +492,123 @@ where
     };
 }
 
+/// Serializes `entry` into `bytes` starting from the index `start`
+/// in `bytes`.
+/// If serialized `entry` is saved in `bytes` on indices `[start, end)`
```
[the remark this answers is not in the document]
dragoljub-duric: It does.

src/storable.rs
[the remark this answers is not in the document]
ielashi: That understanding is out of date. We currently do not store each element with its maximum size. The smaller the element, the smaller its memory footprint - that was the main selling point of the new BTreeMap version that we released, and that's why this optimization can be very impactful.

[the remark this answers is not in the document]
ielashi: @dragoljub-duric I think I misunderstood your last comment. In principle we don't actually I agree. We have that distinction for historical reasons. If we can use the same code for both cases, that would be the ideal.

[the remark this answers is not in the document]
dragoljub-duric: PTAL. If this implementation is fine for you then an implementation for the Unbounded case will use the same code.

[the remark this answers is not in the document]
ielashi: It does. This change should be very simple, right? If yes, it'd be great to include it in this PR so that we see the holistic picture.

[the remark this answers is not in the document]
dragoljub-duric: I have added encoding/decoding for unbounded types. PTAL.

src/storable.rs
```diff
@@ -492,6 +492,120 @@ where
...
+{
+    match T::BOUND {
+        Bound::Bounded {
+            max_size,
+            is_fixed_size,
+        } => bytes_to_store_size(&Bounds {
+            max_size,
+            is_fixed_size,
+        }) as usize,
+        Bound::Unbounded => 8,
```
ielashi, line 507: (communication) why are we returning `8` here? As a reader this isn't clear to me.

src/storable.rs
```diff
@@ -492,6 +492,120 @@ where
     };
 }
 
+fn get_size_len<T>() -> usize
```
ielashi, line 495: (communication) This method's name doesn't explain to me at all what it's for, and what it's intended purpose is. Please pay attention to these names. It makes much more readable and maintainable.

Possible suggestion:

```suggestion
fn get_num_bytes_required_to_store<T>() -> usize
```

src/storable.rs
```diff
@@ -492,6 +492,120 @@ where
...
+    bytes[start + size_len..start + size_len + actual_size].copy_from_slice(entry_bytes);
+
+    start + actual_size + size_len
+}
+
+// Deserialize the struct starting at index `start` in `bytes`.
+// When serialized struct is saved in `bytes` on indices `[start, end)` the
+// function will return deserialized struct and index `end` - the first index
+// after `start` that is not occupied with the serialization of the struct.
+fn deserialize_bounded_with_size<T>(bytes: &[u8], start: usize) -> (T, usize)
```
ielashi, line 534: This method's name is also misleading - it isn't specific to bounded types anymore.

src/storable.rs
```diff
@@ -594,6 +708,54 @@ fn encode_size(dst: &mut [u8], n: usize, bounds: &Bounds) {
     }
 }
 
+fn decode_size_universal<T>(src: &[u8]) -> usize
```
ielashi, line 711: (communication) Again, related to method names. The term "universal" is not clear, and raises more questions than answers. Consider renaming this method to just `decode_size` and rename the current `decode_size` to `decode_size_of_bound`.

src/storable.rs
```diff
@@ -594,6 +708,54 @@ fn encode_size(dst: &mut [u8], n: usize, bounds: &Bounds) {
...
+            debug_assert!(size <= max_size as usize);
+            size
+        }
+        Bound::Unbounded => u64::from_be_bytes([
+            src[0], src[1], src[2], src[3], src[4], src[5], src[6], src[7],
+        ]) as usize,
+    }
+}
+
+fn encode_size_universal<T>(dst: &mut [u8], n: usize)
```
ielashi, line 736: ditto

src/storable.rs
```diff
@@ -492,6 +495,105 @@ where
...
+    let size_len = get_num_bytes_required_to_store_size::<T>();
+    let actual_size = decode_size::<T>(&bytes[start..start + size_len]);
+
+    let a = T::from_bytes(Cow::Borrowed(
+        &bytes[start + size_len..start + size_len + actual_size],
+    ));
+    (a, start + actual_size + size_len)
+}
+
+impl<A, B, C> Storable for (A, B, C)
```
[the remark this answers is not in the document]
dragoljub-duric: 1. TBH, I do not buy this argument. If a developer wants to change Bound from Bounded to Unbounded, that is a different type, so the developer should make it backward compatible. So it can load type as Bounded convert it to Unbounded and save it as Unbounded.
Even more, if we want to make it work in that case we will do even worse. There is no way we can support such behavior with the current serialization of the tuple (A, B) when A and B are bounded because we rely on the Bounds of A and B to determine size. So we will be left with inconsistent behavior because changing the bound will be allowed for (A, B, C) but not for (A, B).
2. I do not think such an operation will be useful on tuples that way, with that exact encoding, counter-example:
   A1 = "10", B1 = "1", A2 = "1", B2 = "11"
   so if we have tuples (A1, B1), (A2, B2)
   and we compare them (using >), expecting that we first compare the first element of the tuple followed by the second element first tuple will be bigger since A1 > A2, but (A1, B1).to_bytes() < (A2, B2).to_bytes() since the second bit in first tuple is "0" while in second is "1" and they have the same size.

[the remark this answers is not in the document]
ielashi: > TBH, I do not buy this argument. If a developer wants to change Bound from Bounded to Unbounded, that is a different type.

This is already a reality. The new btreemap was designed in such a way that allow developers to seamlessly upgrade and remove the bound on their types, and this is something we already support. If we now decide to not support this, we basically tell developers that they can't remove the bounds from their types and will have to create a new btreemap and migrate their data - it's too late to change that (nor is that best for a developer experience). I agree two-element tuples don't support it in its current form, but we can address that issue separately.

> I do not think such an operation will be useful on tuples that way, with that exact encoding, counter-example:
A1 = "10", B1 = "1", A2 = "1", B2 = "11"
so if we have tuples (A1, B1), (A2, B2)

That's a good counter-example. Perhaps there isn't a way to preserve that byte-sorting property in the encoding schema.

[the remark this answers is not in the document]
dragoljub-duric: This is implemented in [PR](https://github.com/dfinity/stable-structures/pull/202).

dragoljub-duric:
This PR is closed in favor of [PR](https://github.com/dfinity/stable-structures/pull/202).

Issue #169, opened 2024-01-10
clear() takes ownership, so it can't be wrapped


borovan:
Hi there, this is our Store which uses a Wrapped stable BTreeMap

![image](https://github.com/dfinity/stable-structures/assets/84897664/94a9214c-a15a-4001-8f94-39e4b816a9b3)

I wanted to wrap the clear() method but I'm unable to, so the workaround is just to go over every key and delete it.  We've got tests with 10,000 inserts so this doesn't seem like a great way to go about it.

I've tried std::mem::take, but that doesn't work as there's no Default value for the map.  There's not much I can do because the MemoryID is required and I can't keep a reference to it for some reason.

Any ideas on how to implement clear?

ielashi:
Hi @borovan, thanks for bringing this up. I think it should be possible to implement `clear` without taking ownership. Let me take a look.

dragoljub-duric:
@borovan Thank you for the request, [here](https://github.com/dfinity/stable-structures/pull/198) is the PR that added the new method `clear_new()` which is in the new [0.6.3](https://github.com/dfinity/stable-structures/releases/tag/v0.6.3) release of Stable Structures, while `clear` is marked as deprecated.

borovan:
Awesome, thanks for this!

PR #198, opened 2024-02-21
chore: Reimplement BTreeMap::clear() to not consume self

dragoljub-duric:
Addresses issue #169


src/btreemap/allocator.rs
```diff
@@ -401,6 +413,40 @@ mod test {
...
+                + AllocatorHeader::size()
+                + allocator.chunk_size()
+                + allocator.chunk_size()
+        );
+        allocator.clear();
+        assert_eq!(
+            allocator.free_list_head,
+            allocator_addr + AllocatorHeader::size()
+        );
+        assert_eq!(allocator.num_allocated_chunks, 0);
```
ielashi, line 439: What if we add new fields to the `Allocator` that need to also be reset on `clear`? If we forget to reset these fields, this test will still pass.

A more robust assertion would be to:
1. Read the `AllocatorHeader`
2. Create a `new` allocator and read its `AllocatorHeader`
3. Verify that headers in steps 1 and 2 are identical.

src/btreemap.rs
```diff
@@ -3029,4 +3037,24 @@ mod test {
...
+            [1u8; 4].as_slice().try_into().unwrap(),
+        );
+
+        assert_ne!(btree.len(), 0);
+        assert_ne!(btree.allocator.num_allocated_chunks(), 0);
+        assert_ne!(btree.root_addr, NULL);
+
+        btree.clear_new();
+
+        assert_eq!(btree.len(), 0);
```
ielashi, line 3056: ditto (regarding testing the headers are the same)

dragoljub-duric: Done, PTAL.

src/btreemap/allocator.rs
```diff
@@ -401,6 +414,35 @@ mod test {
         assert_eq!(allocator.num_allocated_chunks, 0);
     }
 
+    #[test]
+    fn allocate_then_clear() {
```
ielashi, line 418: nit:
```suggestion
    fn clear_deallocates_all_allocated_chunks {
```

src/btreemap/allocator.rs
```diff
@@ -94,6 +95,18 @@ impl<M: Memory> Allocator<M> {
         allocator
     }
 
+    /// Deallocate all allocated chunks.
+    pub fn clear(&mut self) {
```
ielashi, line 99: There is duplicate logic between this function and `fn new()`. Maybe we can merge them together?

```
    pub fn new(memory: M, addr: Address, allocation_size: Bytes) -> Self {
        let mut allocator = Self {
            header_addr: addr,
            allocation_size,
            num_allocated_chunks: 0,
            free_list_head: NULL, // Will be set in `allocator.clear()` below.
            memory,
        };

        allocator.clear();
        allocator
    }

```

src/btreemap.rs
```diff
@@ -3029,4 +3039,28 @@ mod test {
         let btree: BTreeMap<T, T, _> = BTreeMap::init(btree.into_memory());
         assert_eq!(btree.get(&T), Some(T));
     }
+
+    #[test]
+    fn test_clear_new() {
```
ielashi, line 3044: For extra sanity, it would be useful to also have a version of this test where we declare a `BTreeMap` with an unbounded type.

ielashi (approved):

PR #202, opened 2024-03-05
feat: Implement Storable for three elements tuples

dragoljub-duric:


src/storable.rs
```diff
@@ -568,40 +571,192 @@ pub(crate) const fn bounds<A: Storable>() -> Bounds {
...
+
+    const BOUND: Bound = {
+        match (A::BOUND, B::BOUND, C::BOUND) {
+            (Bound::Bounded { .. }, Bound::Bounded { .. }, Bound::Bounded { .. }) => {
+                let a_bounds = bounds::<A>();
+                let b_bounds = bounds::<B>();
+                let c_bounds = bounds::<C>();
+
+                Bound::Bounded {
+                    max_size: a_bounds.max_size
```
ielashi, line 748: Is this `max_size` correct? Aren't we also storing an additional byte for the sizes? Also, we're not storing the size of `C` anymore. We should have test cases that can detect these errors as well.

src/storable.rs
```diff
@@ -568,40 +571,192 @@ pub(crate) const fn bounds<A: Storable>() -> Bounds {
...
+    bytes_to_store_sizes
+}
+
+impl<A, B, C> Storable for (A, B, C)
+where
+    A: Storable,
+    B: Storable,
+    C: Storable,
+{
+    // Touple will be serialized in the following form:
```
ielashi, line 676:
```suggestion
    // Tuple will be serialized in the following form:
```

src/storable.rs
```diff
@@ -568,40 +571,192 @@ pub(crate) const fn bounds<A: Storable>() -> Bounds {
...
+    // <size_lengths (1B)> <size_a (1-4B)> <a_bytes> <size_b(1-4B)> <b_bytes> <c_bytes>
```
ielashi, line 677: If both `A` and `B` are fixed in size, then we don't need to store a size_lengths byte at all, right?

dragoljub-duric: Good point, the code is updated, PTAL.

ielashi:
I think this code can be simplified in a way that makes it easier to read and maintain. Try introducing the following abstractions and see if they help:

```
// Encodes a serialized element `T` in a tuple.
// The element is assumed to be at the beginning of `dst`.
// Returns the number of bytes written to `dst`.
fn encode_tuple_element<T: Storable>(dst: &mut [u8], bytes: Cow<[u8]>, last: bool) -> usize;

// Decodes an element `T` from a tuple.
//
// The element is assumed to be at the beginning of `src`.
// The length of the size of the element should be provided if the element is *not* fixed in size.
//
// Returns the element `T` and the number of bytes read from `src`.
fn decode_tuple_element<T: Storable>(src: &[u8], size_len: Option<u8>, last: bool) -> (T, usize);
```

src/storable.rs
ielashi: Can we also add tests that also assert the max size to prevent this regression?

ielashi (commented):
Code looks good! I left one minor comment about adding tests for the max size.

By the way, I think in a followup PR it would be useful to move all the code related to encoding tuples to a separate file since it's quite a lot (e.g. `src/storable/tuples.rs`) to keep things a bit more organized. WDYT?

dragoljub-duric:
@ielashi Bound tests are added, PTAL. I agree in the follow up we can move parts to `src/storable/tuples.rs`.

ielashi (approved):
LGTM, thanks!

PR #212, opened 2024-04-17
fix: BTreeMap memory leak when deallocating nodes with overflows

ielashi:
# Problem
When storing unbounded types in a `BTreeMap`, a node is represented as a linked list of "memory chunks". It was discovered recently that when we deallocate a node, in some cases only the first memory chunk is deallocated, and the rest of the memory chunks remain (incorrectly) allocated, causing a memory leak.

# Solution
Change the logic for deallocating nodes to ensure that all of a node's memory chunks are deallocated. Tests have been added to prevent regressions of this nature moving forward.


dsarlis (approved):

dsarlis (commented):
Did you also consider extending some of the benchmarks we have to ensure that we don't see any increase in stable memory usage during a `remove` operation?

ielashi:
> Did you also consider extending some of the benchmarks we have to ensure that we don't see any increase in stable memory usage during a `remove` operation?

In the current implementation we do increase stable memory in some cases on `remove`. This PR does not address that, and strictly speaking it's not a bug. I'll be looking into how/whether we can avoid that, but for now I'm prioritizing resolving the memory leak.

PR #222, opened 2024-06-07
fix: grow memory when necessary in `BaseVec::set`.

ielashi:
## Problem

`BaseVec`, which is the underlying data structure used by `StableVec` and `StableMinHeap` provides a `set` method that allows the caller to change the value of an element at a specified index.

The `set` method used the `memory.write` method, which assumed that there is always enough memory to rewrite the element. This assumption would've been fine if other methods like `push` always allocated the maximum amount of space for every element. Currently though, `push` only allocates the amount needed to store the element that's being pushed.

## Solution

There are two possible solutions:

1. Modify `push` to always allocate the maximum size of an element.
2. Modify `set` to grow the memory as needed as it's rewriting an element.

This commit implements solution 2, which is both simpler and more performant.

### DIFF
--- a/src/base_vec.rs
+++ b/src/base_vec.rs
@@ -33,7 +33,8 @@
 //! bytes required to represent integers up to that max size.
 use crate::storable::{bounds, bytes_to_store_size_bounded};
 use crate::{
-    read_u32, read_u64, safe_write, write_u32, write_u64, Address, GrowFailed, Memory, Storable,
+    read_u32, read_u64, safe_write, write, write_u32, write_u64, Address, GrowFailed, Memory,
+    Storable,
 };
 use std::borrow::{Borrow, Cow};
 use std::cmp::min;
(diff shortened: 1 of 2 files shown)


src/base_vec.rs
```diff
@@ -183,7 +184,7 @@ impl<T: Storable, M: Memory> BaseVec<T, M> {
         let data_offset = self
             .write_entry_size(offset, bytes.len() as u32)
             .expect("unreachable: cannot fail to write to pre-allocated area");
-        self.memory.write(data_offset, bytes.borrow());
+        write(&self.memory, data_offset, bytes.borrow());
```
ielashi, line 187: Note: if we want to be consistent with other APIs in this data structure, we would've used `safe_write` here and then `set` would return a `Result<(), GrowFailed>`. This would be a breaking change that would affect both `StableVec` and `StableMinHeap`, and I didn't feel like it was worth the effort/hassle given that the benefit of returning a `GrowFailed` error is still questionable. In fact, I'm wondering if we should remove the `GrowFailed` error from all the other APIs given that it hasn't been that useful to my knowledge, and just makes code more verbose.

dsarlis: I'd be in favour of removing as I'm also not sure what can a canister reasonably do if it hits the `GrowFailed` case. What course of action would you take? If you were able to run some logic when getting close to out of memory (aka having some hook) that could be useful but if you hit this error, you might as well trap yourself.

dsarlis (approved):

src/base_vec.rs
ielashi: Exactly. I think if the course of action to take would be, for example, spawn another canister to move your data, you'd want to do it well before you hit the grow limit. I'll create a ticket, and we can take care of this separately.

PR #231, opened 2024-10-16
feat: move initialization of `StableBTreeMap` cursors into `Iter`

hpeebles:


src/btreemap/iter.rs
```diff
@@ -29,6 +29,9 @@ where
     // A reference to the map being iterated on.
     map: &'a BTreeMap<K, V, M>,
 
+    // A flag indicating if the cursors have been initialized yet.
+    cursors_initialized: bool,
```
dsarlis, line 33: I'm not sure I understand the point of this flag. Conceptually, it seems that it should be tied with whether the `vec` of cursors below is empty or not but there's this case where we set it to true for the `null` constructor. Can you maybe elaborate on the motivation for adding this?

dsarlis (commented):
Let's add a short description of what the changes are trying to accomplish.

I also notice a non-negligible performance regression for `insert` operations. Do you have an idea why that might be?

src/btreemap/iter.rs
hpeebles: Prior to this PR, the cursors would be calculated before constructing each `Iter` instance. Then once iteration was finished the `cursors` field would be an empty `Vec`.

As of this PR, the calculation of the cursors is deferred until the first call to `Iterator::next`, so there needs to be a way to distinguish between the case where the iteration hasn't started yet and the case where the iteration has finished (at which point the cursors field is empty.

So I added the `cursors_initialized` field to signal if the cursors have been calculated yet, making it possible to tell if iteration is yet to start or if it has already finished.

In my follow on PR there will be 2 fields `forward_cursors_initialized` and `backward_cursors_initialized` because the cursors for each direction will be calculated lazily only if required.

hpeebles:
> Let's add a short description of what the changes are trying to accomplish.
> 
> I also notice a non-negligible performance regression for `insert` operations. Do you have an idea why that might be?

I'm surprised by the performance regression around inserts since I should have only affected reading and iterating over the map. I'll have a look into it tomorrow though.

src/btreemap/iter.rs
dsarlis: I see, thanks for the explanation.

> the calculation of the cursors is deferred until the first call to Iterator::next, so there needs to be a way to distinguish between the case where the iteration hasn't started yet and the case where the iteration has finished (at which point the cursors field is empty.

This or (a more succinct) part of it could be added in the comment for the flag I think.

dsarlis:
> > Let's add a short description of what the changes are trying to accomplish.
> > I also notice a non-negligible performance regression for `insert` operations. Do you have an idea why that might be?
> 
> I'm surprised by the performance regression around inserts since I should have only affected reading and iterating over the map. I'll have a look into it tomorrow though.

`insert` also needs to navigate the map to find the point where the value should be inserted, so if your change affects iteration (in more general terms "navigation" in the map), it's not completely unexpected to affect `insert` I think.

dsarlis:
> > > Let's add a short description of what the changes are trying to accomplish.
> > > I also notice a non-negligible performance regression for `insert` operations. Do you have an idea why that might be?
> > 
> > 
> > I'm surprised by the performance regression around inserts since I should have only affected reading and iterating over the map. I'll have a look into it tomorrow though.
> 
> `insert` also needs to navigate the map to find the point where the value should be inserted, so if your change affects iteration (in more general terms "navigation" in the map), it's not completely unexpected to affect `insert` I think.

Actually, I think we might be seeing a race condition here because around the same time, Islam [bumped the canbench version used](https://github.com/dfinity/stable-structures/pull/232). By visually comparing the numbers, I think you get pretty much the same after that PR.

src/btreemap/iter.rs
hpeebles: I've added a comment to the `cursors_initialized` field

dsarlis (approved):

dsarlis (approved):

PR #235, opened 2024-11-07
feat: implement `DoubleEndedIterator` for `StableBTreeMap`

hpeebles:
This adds the ability to iterate a `StableBTreeMap` in either direction.
We have been using this within OpenChat for a few weeks now with no issues.

### DIFF
--- a/src/btreemap/iter.rs
+++ b/src/btreemap/iter.rs
@@ -335,7 +335,9 @@ where
                     NodeType::Internal => Index::Child(entry_idx + 1),
                     // If this is a leaf node, and it contains another entry, add the
                     // next entry to the cursors.
-                    NodeType::Leaf if entry_idx + 1 < node.entries_len() => Index::Entry(entry_idx + 1),
+                    NodeType::Leaf if entry_idx + 1 < node.entries_len() => {
+                        Index::Entry(entry_idx + 1)
+                    }
                     _ => return Some(res),
                 };
 
--- a/src/btreemap/iter.rs
+++ b/src/btreemap/iter.rs
@@ -83,22 +83,6 @@ where
         }
     }
 
-    // This can be used as an optimisation if the cursors have already been calculated
-    pub(crate) fn new_with_cursors(
-        map: &'a BTreeMap<K, V, M>,
-        range: (Bound<K>, Bound<K>),
-        cursors: Vec<Cursor<K>>,
-    ) -> Self {
-        Self {
-            map,
-            forward_cursors_initialized: true,
-            backward_cursors_initialized: false,
-            forward_cursors: cursors,
-            backward_cursors: vec![],
-            range,
-        }
-    }
-
     fn initialize_forward_cursors(&mut self) {
         debug_assert!(!self.forward_cursors_initialized);
 
--- a/src/btreemap/iter.rs
+++ b/src/btreemap/iter.rs
@@ -88,8 +88,7 @@ where
 
         match self.range.start_bound() {
             Bound::Unbounded => {
-                self.forward_cursors
-                    .push(Cursor::Address(self.map.root_addr));
+                self.forward_cursors.push(Cursor::Address(self.map.root_addr));
             }
             Bound::Included(key) | Bound::Excluded(key) => {
                 let mut node = self.map.load_node(self.map.root_addr);
@@ -182,8 +181,7 @@ where
 
         match self.range.end_bound() {
             Bound::Unbounded => {
-                self.backward_cursors
-                    .push(Cursor::Address(self.map.root_addr));
+                self.backward_cursors.push(Cursor::Address(self.map.root_addr));
             }
             Bound::Included(key) | Bound::Excluded(key) => {
                 let mut node = self.map.load_node(self.map.root_addr);
@@ -208,7 +206,8 @@ where
                                     NodeType::Leaf => None,
                                 };
 
-                                if idx > 0 && self.range.contains(node.key(idx - 1)) {
+                                if idx > 0 && self.range.contains(node.key(idx - 1))
+                                {
                                     self.backward_cursors.push(Cursor::Node {
                                         node,
                                         next: Index::Entry(idx - 1),
@@ -364,16 +363,15 @@ where
                     let node = self.map.load_node(address);
                     if let Some(next) = match node.node_type() {
                         // Iterate on internal nodes starting from the first child.
-                        NodeType::Internal if node.children_len() > 0 => {
-                            Some(Index::Child(node.children_len() - 1))
-                        }
+                        NodeType::Internal if node.children_len() > 0 => Some(Index::Child(node.children_len() - 1)),
                         // Iterate on leaf nodes starting from the first entry.
-                        NodeType::Leaf if node.entries_len() > 0 => {
-                            Some(Index::Entry(node.entries_len() - 1))
-                        }
-                        _ => None,
+                        NodeType::Leaf if node.entries_len() > 0 => Some(Index::Entry(node.entries_len() - 1)),
+                        _ => None
                     } {
-                        self.backward_cursors.push(Cursor::Node { next, node });
+                        self.backward_cursors.push(Cursor::Node {
+                            next,
+                            node,
+                        });
                     }
                 }
                 self.next_back_map(map)
@@ -419,10 +417,13 @@ where
                     NodeType::Internal => Some(Index::Child(entry_idx)),
                     // If this is a leaf node, add the previous entry to the cursors.
                     NodeType::Leaf if entry_idx > 0 => Some(Index::Entry(entry_idx - 1)),
-                    _ => None,
+                    _ => None
                 } {
                     // Add to the cursors the next element to be traversed.
-                    self.backward_cursors.push(Cursor::Node { next, node });
+                    self.backward_cursors.push(Cursor::Node {
+                        next,
+                        node,
+                    });
                 }
 
                 Some(res)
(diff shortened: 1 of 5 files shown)


hpeebles:
I don't really understand the use case for `iter_upper_bound`, but it previously was calculating the cursors to start iterating from, whereas now that we can iterate backwards we can simply use `next_back` to calculate the starting point and then iterate forwards from there.
This does unnecessarily deserialize the value at the starting point, so we could potentially add something like `iter_keys` and `range_keys` which does the same thing but only yields the keys without deserializing the values.

src/btreemap/iter.rs
```diff
@@ -239,14 +325,15 @@ where
                 // If the key does not belong to the range, iteration stops.
                 if !self.range.contains(node.key(entry_idx)) {
                     // Clear all cursors to avoid needless work in subsequent calls.
-                    self.cursors = vec![];
+                    self.forward_cursors = vec![];
                     return None;
                 }
 
                 let res = map(&node, entry_idx);
+                self.range.0 = Bound::Excluded(node.key(entry_idx).clone());
```
hpeebles, line 333: Updating the remaining range causes a small regression in the benchmarks. If I comment out this line everything stays a 0% change. But this is required in order to iterate backwards up to the correct bound.
I think in a real canister this would be reading keys from stable memory so the impact of setting this field on the heap will become negligible.

src/btreemap/iter.rs
```diff
@@ -239,14 +326,15 @@ where
                 // If the key does not belong to the range, iteration stops.
                 if !self.range.contains(node.key(entry_idx)) {
                     // Clear all cursors to avoid needless work in subsequent calls.
-                    self.cursors = vec![];
+                    self.forward_cursors = vec![];
```
dsarlis, line 329: Shouldn't we also clear `backward_cursors` here as well?

src/btreemap/iter.rs
```diff
@@ -338,4 +521,119 @@ mod test {
...
+    fn iterate_reverse() {
+        let mem = make_memory();
+        let mut btree = BTreeMap::new(mem);
+
+        // Insert the elements in reverse order.
+        for i in (0..100u64).rev() {
+            btree.insert(i, i + 1);
+        }
+
+        // Iteration should be in ascending order.
```
dsarlis, line 556:
```suggestion
        // Iteration should be in descending order.
```

src/btreemap/iter.rs
```diff
@@ -338,4 +521,119 @@ mod test {
...
+    fn iterate_range_reverse() {
+        let mem = make_memory();
+        let mut btree = BTreeMap::new(mem);
+
+        // Insert the elements in reverse order.
+        for i in (0..100u64).rev() {
+            btree.insert(i, i + 1);
+        }
+
+        // Iteration should be in ascending order.
```
dsarlis, line 577:
```suggestion
        // Iteration should be in descending order.
```

src/btreemap/iter.rs
```diff
@@ -158,10 +153,102 @@ where
...
+                                break;
+                            }
+                        }
+                        Err(idx) => {
+                            // The `idx` variable points to the first
+                            // key that is greater than the left
+                            // bound.
+                            //
+                            // If the index points to a valid node, we
+                            // will visit its left subtree and then
```
dsarlis, line 229:
```suggestion
                            // will visit its right subtree and then
```

src/btreemap/iter.rs
```diff
@@ -158,10 +153,102 @@ where
...
+                            // return to this key.
+                            //
+                            // If the index points at the end of
+                            // array, we'll continue with the right
```
dsarlis, line 233:
```suggestion
                            // array, we'll continue with the left
```

src/btreemap/iter.rs
```diff
@@ -158,10 +153,102 @@ where
...
+                                    Some(self.map.load_node(node.child(idx)))
+                                }
+                                NodeType::Leaf => None,
+                            };
+
+                            if idx > 0 && self.range.contains(node.key(idx - 1)) {
+                                self.backward_cursors.push(Cursor::Node {
+                                    node,
+                                    next: Index::Entry(idx - 1),
+                                });
```
dsarlis, line 251: I'm not entirely sure about the above comments but it feels like we need to adjust something as there's some diff in what we do here vs in the forward cursors case.

src/btreemap/iter.rs
```diff
@@ -260,6 +348,87 @@ where
...
+        }
+
+        // If the cursors are empty. Iteration is complete.
+        match self.backward_cursors.pop()? {
+            Cursor::Address(address) => {
+                if address != NULL {
+                    // Load the node at the given address, and add it to the cursors.
+                    let node = self.map.load_node(address);
+                    if let Some(next) = match node.node_type() {
+                        // Iterate on internal nodes starting from the first child.
```
dsarlis, line 366:
```suggestion
                        // Iterate on internal nodes starting from the last child.
```

src/btreemap/iter.rs
```diff
@@ -260,6 +348,87 @@ where
...
+                        NodeType::Internal if node.children_len() > 0 => {
+                            Some(Index::Child(node.children_len() - 1))
+                        }
+                        // Iterate on leaf nodes starting from the first entry.
```
dsarlis, line 370:
```suggestion
                        // Iterate on leaf nodes starting from the last entry.
```

src/btreemap/iter.rs
```diff
@@ -260,6 +348,87 @@ where
...
+                // Add the child to the top of the cursors to be iterated on first.
+                self.backward_cursors.push(Cursor::Address(child_address));
+
+                self.next_back_map(map)
+            }
+
+            Cursor::Node {
+                node,
+                next: Index::Entry(entry_idx),
+            } => {
```
dsarlis, line 405: Should we also have a short-circuit here like in the `next_map` where we do something like
```
if entry_idx <= 0 {
    // No more entries to iterate on in this node.
    return self.next_map(map);
}
```

I think this might then simplify the code later below because you could just use `entry_idx - 1` without worrying about it and it would be more similar to the other case (which makes it easier to inspect correctness). Unless there's something else I'm missing of course.

dsarlis:
Can you also revert the changes in `canbench_results.txt`, I'd like to first see what's the diff in benches, then we can put it back in (I expect a very small acceptable regression that you also mentioned but would like to confirm).

dsarlis:
> Maybe keys and range_keys are worth adding regardless of this extra deserialization. I'd be happy to add them in a follow on PR if others agree they are worth having.

I don't feel particularly strongly either way. If indeed people find them useful, I don't see why not adding them. In a sense, we should ideally provide a very similar interface to the std BTreeMap but of course these things typically happen on a "as needed basis".

src/btreemap/iter.rs
hpeebles: I just mean that in the benchmarks the memory used is `VectorMemory` which is significantly faster than `Ic0StableMemory`, so the tiny regression in the benchmarks will become even smaller in reality when `Ic0StableMemory` is being used.

src/btreemap/iter.rs
hpeebles: In the `next_map` case, the check on if the index exceeds the entries was never done when pushing cursors, it was then handled by the `if entry_idx >= node.entries_len() { .. }` check.
Whereas in the `next_map_back` the check was being done before pushing the cursor to prevent underflow (entry_id of 0 is valid).
I've now modified it so that the checks always happen before pushing the cursors when iterating in either direction.
This means the check happens in 2 places rather than 1 in the forward case, but it also means we skip pushing and popping unnecessary cursors that we know won't contain any items.

src/btreemap/iter.rs
```diff
@@ -260,6 +348,87 @@ where
...
+                    self.backward_cursors = vec![];
+                    return None;
+                }
+
+                let res = map(&node, entry_idx);
+                self.range.1 = Bound::Excluded(node.key(entry_idx).clone());
+
+                if let Some(next) = match node.node_type() {
+                    // If this is an internal node, add the previous child to the cursors.
+                    NodeType::Internal => Some(Index::Child(entry_idx)),
```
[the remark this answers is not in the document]
hpeebles: The child before each entry has the same index as the entry so I'm fairly sure this is correct.
Also the tests pass like this whereas I've just tried it using `entry_idx - 1` and they start failing.

src/btreemap/iter.rs
dsarlis: I don't think this is accurate. The benchmarks are using canbench which is running real canisters. It wouldn't make much sense to do anything else as we wouldn't be benchmarking what would actually be running in production. Do you have evidence that a `VectorMemory` is used instead or was it an assumption of yours?

src/btreemap/iter.rs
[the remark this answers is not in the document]
dsarlis: You are right, I'd forgotten how the child indices work.

dsarlis (approved):

dsarlis:
@hpeebles I do see some non-negligible regressions in iter benchmarks (both for smaller and larger values). Do you think we could maybe do something to make them a bit better?

src/btreemap/iter.rs
hpeebles: Ah right my mistake then. This is my first time using `canbench` and when I look at the functions it doesn't appear as if they're running within real canisters but I guess there's some magic happening to make them do so.

dsarlis (approved):

dsarlis (approved):

PR #236, opened 2024-11-11
feat: Support memory freeing

hpeebles:
WIP


src/memory_manager.rs
```diff
@@ -294,6 +291,8 @@ impl<M: Memory> Memory for VirtualMemory<M> {
 struct MemoryManagerInner<M: Memory> {
     memory: M,
 
+    version: u8,
```
dragoljub-duric, line 294: Do we need this?

src/memory_manager.rs
```diff
@@ -799,8 +821,144 @@ fn bucket_allocations_address(id: BucketId) -> Address {
...
+        }
+    }
+
+    // Flushes all bytes to memory
+    fn flush_all<M: Memory>(&mut self, memory: &M, start_offset: u64) {
+        write_struct(&self.inner, Address::from(start_offset), memory);
+        self.dirty_first_buckets.clear();
+        self.dirty_bucket_link_bytes.clear();
+    }
+}
```
dragoljub-duric, line 943: I would strongly prefer to have only one place where we read the header, and after that, we use u16 (instead of 15-bit representation in storage). I believe that will improve performance, but more importantly, it will be easier to maintain and understand. After that, we can still use information from dirty_first_buckets and dirty_bucket_link_bytes to apply changes in storage on 15-bit representation when saving.

src/memory_manager.rs
hpeebles: It is only used in the `upgrade_from_v1_to_v2` test to ensure the correct version of `grow` is called.
I'll wrap it in a `cfg(test)`.

src/memory_manager.rs
hpeebles: Yeah that makes sense.
This means we'll have to convert from 15 to 16-bit after reading which will have some overhead but at least its only done once. I'll have a go at it now.

hpeebles: Done. Will likely need a bit of cleaning up. Need to run off now but will give it another look tomorrow.

hpeebles:
The vast majority of use cases won't need a huge number of buckets.
So we could keep the header as a single page, store each bucket in the linked list as 2 bytes, and if additional buckets are needed, we use the unused space in the header to point to where the subsequent buckets are defined.
Maybe these follow on buckets are defined as 3 bytes, and at the end of the page is a pointer to the next page of buckets, and so on.
That way the number of buckets is pretty much unbounded, the header remains small in most use cases, and we wouldn't need to move the existing data in the first memory page after the header.

hpeebles:
Having thought about this a bit more I think my idea to store bucketIds as 3 bytes is wrong.
If we support spreading buckets onto more pages, then storing them as 3 bytes is an unnecessary storage optimisation at the expense of code complexity. I think storing buckets as u32s is the way to go.
The first page could then store something like 8192 buckets in the last 32768 bytes (using the format I introduced in this PR where each bucket points to the following bucket in the linked list).
Follow on pages used to store buckets could then store 16384 per page.
The first page then just needs to somehow store where the additional pages are.

PR #237, opened 2024-11-11
Optimize how values are stored and handled

frankdavid:
This PR contains 2 optimizations:
- Stores keys and values together in a single `Vec` instead of in separated `Vec`s
- Reuses values in methods that remove an entry from a `Node`. These methods previously fetched the value using `.value()` and discarded the value returned during the removal which was wasteful.

### DIFF
--- a/src/btreemap/node.rs
+++ b/src/btreemap/node.rs
@@ -49,10 +49,11 @@ pub type Entry<K> = (K, Vec<u8>);
 #[derive(Debug)]
 pub struct Node<K: Storable + Ord + Clone> {
     address: Address,
-    keys: Vec<K>,
+    // List of tuples consisting of a key and the encoded value.
     // Values are stored in a Refcell as they are loaded lazily.
     // A RefCell allows loading the value and caching it without requiring exterior mutability.
-    encoded_values: RefCell<Vec<Value>>,
+    // INVARIANT: the list is sorted by key.
+    keys_encoded_values: Vec<(K, RefCell<Value>)>,
     // For the key at position I, children[I] points to the left
     // child of this key and children[I + 1] points to the right child.
     children: Vec<Address>,
@@ -106,9 +107,13 @@ impl<K: Storable + Ord + Clone> Node<K> {
     pub fn get_max<M: Memory>(&self, memory: &M) -> Entry<K> {
         match self.node_type {
             NodeType::Leaf => {
-                let last_idx = self.encoded_values.borrow().len() - 1;
+                let last_idx = self.keys_encoded_values.len() - 1;
                 (
-                    self.keys.last().expect("A node can never be empty").clone(),
+                    self.keys_encoded_values
+                        .last()
+                        .expect("A node can never be empty")
+                        .0
+                        .clone(),
                     self.value(last_idx, memory).to_vec(),
                 )
             }
@@ -147,7 +152,7 @@ impl<K: Storable + Ord + Clone> Node<K> {
 
     /// Returns true if the node cannot store anymore entries, false otherwise.
     pub fn is_full(&self) -> bool {
-        self.keys.len() >= CAPACITY
+        self.keys_encoded_values.len() >= CAPACITY
     }
 
     /// Swaps the entry at index `idx` with the given entry, returning the old entry.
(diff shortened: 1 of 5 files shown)


src/btreemap/node.rs
```diff
@@ -49,10 +49,11 @@ pub type Entry<K> = (K, Vec<u8>);
 #[derive(Debug)]
 pub struct Node<K: Storable + Ord + Clone> {
     address: Address,
-    keys: Vec<K>,
+    // List of tuples consisting of a key and the encoded value.
     // Values are stored in a Refcell as they are loaded lazily.
     // A RefCell allows loading the value and caching it without requiring exterior mutability.
-    encoded_values: RefCell<Vec<Value>>,
+    // INVARIANT: the list is sorted by key.
+    keys_encoded_values: Vec<(K, RefCell<Value>)>,
```
dsarlis, line 56: Perhaps `keys_and_encoded_values` is a bit better? It took me a moment there to parse the name (until I wrapped my head around the tuple).

src/btreemap/node.rs
```diff
@@ -147,34 +152,55 @@ impl<K: Storable + Ord + Clone> Node<K> {
...
     pub fn value<M: Memory>(&self, idx: usize, memory: &M) -> Ref<[u8]> {
         // Load and cache the value from the underlying memory if needed.
-        {
-            let mut values = self.encoded_values.borrow_mut();
+        let mut value = self.keys_encoded_values[idx].1.borrow_mut();
+        if let Value::ByRef(offset) = *value {
+            // Cache the value internally.
+            *value = Value::ByVal(self.resolve_value(Value::ByRef(offset), memory));
+        }
+        drop(value); // drop borrow because it's reborrowed below
```
dsarlis, line 189: nit: Could we not use an explicit scope to help the borrow checker? I find that a bit more readable than an explicit `drop` that you need to accompany with the comment. The explicit scope is almost always an indication that you want to force something to be dropped. So, something like:

```
{
   let mut value = self.keys_encoded_values[idx].1.borrow_mut();
   if let Value::ByRef(offset) = *value {
      // Cache the value internally.
      *value = Value::ByVal(self.resolve_value(Value::ByRef(offset), memory));
   }
}
```

src/btreemap/node.rs
```diff
@@ -147,34 +152,55 @@ impl<K: Storable + Ord + Clone> Node<K> {
...
+
+        // Return a reference to the value.
+        Ref::map(self.keys_encoded_values[idx].1.borrow(), |values| {
+            if let Value::ByVal(v) = values {
+                &v[..]
+            } else {
+                unreachable!("value must have been loaded already.");
+            }
+        })
```
dsarlis, line 198: Did you try to avoid the `unreachable` by restructuring the code and handling things in a single match and it didn't work? Is that why the above code snippet as well? It seems like it (I guess we cannot be iterating over a collection and wanting to change it at the same time). In that case, I suggest the following:

1. Expand the error message a bit to say "have been loaded already a few lines above" or so.
2. Maybe leave a comment that explains that we do this because iterating over the values and wanting to change them at the same time (in case we first want to cache it, then return it) does not work and we are forced to go down the `unreachable` path.

dsarlis: Actually, the more I am reading this code, I don't see why we need the `Ref::map`? Can't we just index in the value we want and return that directly?

src/btreemap/node.rs
frankdavid: Actually, I realized that both the old and new implementation had a bug which made the code panic when we called `.value()` multiple times with the same index. I fixed it and added a test, PTAL.

src/btreemap/node.rs
frankdavid: Note that this unreachable() was there already before, I just moved it. I don't think it's possible to remove it if we want to keep the `Ref<[u8]>`, since referencing the value out of the borrow has to happen in the closure passed to `Ref::map`.

`Ref::map` is necessary if we want to avoid cloning the value (the alternative would be returning a `Vec<u8>` and cloning on every invocation). Without `Ref` the borrow would die when the function returns so the returned reference would be invalid.

frankdavid: One more alternative in the future would be calling `Ref::leak` which would return an immutable naked reference with the lifetime of the Node. However, this is only available on nightly at the moment.

frankdavid: One more alternative would be using `OnceCell` (it was stabilized in Rust 1.80). It would require some refactoring, though. With `OnceCell` we could avoid `unreachable!()` and it would be perhaps a little more expressive than `RefCell`.

dsarlis: Thanks for the detailed explanation!

Yeah I know it was there before, I was mainly asking to see if there's anything possible now that you are refactoring some of the code. I believe we can simply keep the unreachable then, no need to get a lot more fancy.

dsarlis (approved):

src/btreemap/node.rs
frankdavid: Improved the error message and replaced the code inside `Ref::map` to use a `match` with explicit arms. This way we'll get a compile-time error message if we ever add new enum variants to `Value`.

dsarlis (dismissed):

PR #240, opened 2024-11-12
perf: Replace BTreeMap with constant vector in MemoryManager

frankdavid:
Since we can have maximum 255 memories, we can store those cheaply in a `Vec` and save time during the lookup.

### DIFF
--- a/src/btreemap/node.rs
+++ b/src/btreemap/node.rs
@@ -49,10 +49,11 @@ pub type Entry<K> = (K, Vec<u8>);
 #[derive(Debug)]
 pub struct Node<K: Storable + Ord + Clone> {
     address: Address,
-    keys: Vec<K>,
+    // List of tuples consisting of a key and the encoded value.
     // Values are stored in a Refcell as they are loaded lazily.
     // A RefCell allows loading the value and caching it without requiring exterior mutability.
-    encoded_values: RefCell<Vec<Value>>,
+    // INVARIANT: the list is sorted by key.
+    keys_encoded_values: Vec<(K, RefCell<Value>)>,
     // For the key at position I, children[I] points to the left
     // child of this key and children[I + 1] points to the right child.
     children: Vec<Address>,
@@ -106,9 +107,13 @@ impl<K: Storable + Ord + Clone> Node<K> {
     pub fn get_max<M: Memory>(&self, memory: &M) -> Entry<K> {
         match self.node_type {
             NodeType::Leaf => {
-                let last_idx = self.encoded_values.borrow().len() - 1;
+                let last_idx = self.keys_encoded_values.len() - 1;
                 (
-                    self.keys.last().expect("A node can never be empty").clone(),
+                    self.keys_encoded_values
+                        .last()
+                        .expect("A node can never be empty")
+                        .0
+                        .clone(),
                     self.value(last_idx, memory).to_vec(),
                 )
             }
@@ -147,7 +152,7 @@ impl<K: Storable + Ord + Clone> Node<K> {
 
     /// Returns true if the node cannot store anymore entries, false otherwise.
     pub fn is_full(&self) -> bool {
-        self.keys.len() >= CAPACITY
+        self.keys_encoded_values.len() >= CAPACITY
     }
 
     /// Swaps the entry at index `idx` with the given entry, returning the old entry.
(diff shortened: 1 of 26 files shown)


dsarlis:
I am not entirely sure about this change. It works well maybe now that we can only allocate new memories but what if we finally merge the change which allows to free a memory? Wouldn't that nullify the benefits of using a vector?

frankdavid:
> Wouldn't that nullify the benefits of using a vector?

If you think we'd need to delete elements from the middle of the vector, that's not the case. We'll never delete elements from the outer `Vec` in `Vec<Vec<BucketId>>`. The outer `Vec` always has 255 elements (it could also be a fixed-size array, I'm using a `Vec` to keep the struct's size relatively low, thus cheap to move). When a memory is empty, the corresponding inner `Vec` will be empty. For example if memory with id 32 is empty, `memories[32].is_empty()` returns true.

We can remove a memory by simply taking its contents, eg:
`std::mem::take(memories[32])` which returns the `Vec` of `BucketId`s that were in the memory and leaves an empty `Vec` behind. Or we can call `memories[32].clear()` to simply clear all buckets from it.

If this is confusing to read, I could wrap the whole thing into a struct which would have a similar API like stdlib `BTreeMap` but with the performance benefits of a linear `Vec`. I think the performance benefits are worth it - in multiple benchmarks it's 4-7% improvement and it's a fairly trivial change.

dsarlis:
> > Wouldn't that nullify the benefits of using a vector?
> 
> If you think we'd need to delete elements from the middle of the vector, that's not the case. We'll never delete elements from the outer `Vec` in `Vec<Vec<BucketId>>`. The outer `Vec` always has 255 elements (it could also be a fixed-size array, I'm using a `Vec` to keep the struct's size relatively low, thus cheap to move). When a memory is empty, the corresponding inner `Vec` will be empty. For example if memory with id 32 is empty, `memories[32].is_empty()` returns true.
> 
> We can remove a memory by simply taking its contents, eg: `std::mem::take(memories[32])` which returns the `Vec` of `BucketId`s that were in the memory and leaves an empty `Vec` behind. Or we can call `memories[32].clear()` to simply clear all buckets from it.
> 
> If this is confusing to read, I could wrap the whole thing into a struct which would have a similar API like stdlib `BTreeMap` but with the performance benefits of a linear `Vec`. I think the performance benefits are worth it - in multiple benchmarks it's 4-7% improvement and it's a fairly trivial change.

I see, yeah I was thinking we might need to remove in the middle of the vector but I get your point. In this case, I agree the change is beneficial.

dragoljub-duric:
@frankdavid nice change, thank you. I approve this change, but can we please wait for the memory freeing to first be merged, we have a couple of stacked PRs, and I hope that can be done next week. And after that, I will put a stamp here. Does that sound fine?

frankdavid:
Sure, no problem.

dragoljub-duric:
Hey @frankdavid just an update, I believe the work on the memory manager will take more time than initially assumed, so please go forward and merge this PR if you think it is ready.

dragoljub-duric (dismissed):

src/btreemap/node.rs
```diff
@@ -49,10 +49,11 @@ pub type Entry<K> = (K, Vec<u8>);
 #[derive(Debug)]
 pub struct Node<K: Storable + Ord + Clone> {
     address: Address,
-    keys: Vec<K>,
+    // List of tuples consisting of a key and the encoded value.
     // Values are stored in a Refcell as they are loaded lazily.
     // A RefCell allows loading the value and caching it without requiring exterior mutability.
-    encoded_values: RefCell<Vec<Value>>,
+    // INVARIANT: the list is sorted by key.
+    keys_encoded_values: Vec<(K, RefCell<Value>)>,
```
dragoljub-duric, line 56: Is this required in order to change MemoryManager?

frankdavid: I merged in master, should be resolved.

frankdavid:
After merging recent changes from master, I no longer see significant performance improvements from this change, so I won't move forward with this PR for now.

frankdavid:
After studying our code more, I realized that we didn't have any benchmarks that properly exercised the MemoryManager, that's why we didn't see any improvements. A few new benchmarks were added in  #253 and now the effect of this PR is visible.

dragoljub-duric (approved):

PR #241, opened 2024-11-12
feat: implement `keys` and `values` on `StableBTreeMap`

hpeebles:


dsarlis (commented):
I feel like it would be a good idea to introduce some benchmarks that cover the new functions added as well, wdyt?

hpeebles:
> I feel like it would be a good idea to introduce some benchmarks that cover the new functions added as well, wdyt?

Done!

dsarlis (approved):

PR #244, opened 2024-11-18
opt: Replace Vec.resize calls with Vec.reserve + Vec.set_len

frankdavid:
This involves unsafe code, but results in significant performance gains. Since implementations don't read the contents of the buffer (this PR also codifies this assumption in a doc comment), it's safe to skip initializing the buffer before passing it to `Memory.read`.


frankdavid:
This approach is unsound according to Rust rules and I'll send out another approach that is sound.

PR #245, opened 2024-11-19
perf: Refactor the internal Value data structure and make some performance improvements

frankdavid:
This PR refactors the internal Value data structure that is used to hold a loaded or an unloaded map value. Previous we had stored `Rc<Enum>`-s where `Enum` was either a loaded value (`Vec<u8>`) or an unloaded value (`offset: Bytes`). The new representation stores a `OnceCell` inside the enum so we no longer need `Rc`.

The new representation makes it possible to hand out plain references to the loaded values (whereas a special `Ref` wrapper was necessary previously). Using this, we can optimize some of the code paths leading to good performance gains.

### DIFF
--- a/src/btreemap/node.rs
+++ b/src/btreemap/node.rs
@@ -503,8 +503,8 @@ impl Value {
             Value::ByVal(v) => v,
             Value::ByRef {
                 offset,
-                loaded_value: mut value,
-            } => value.take().unwrap_or_else(|| load(offset)),
+                loaded_value: value,
+            } => value.into_inner().unwrap_or_else(|| load(offset)),
         }
     }
 }
--- a/src/btreemap/node.rs
+++ b/src/btreemap/node.rs
@@ -454,7 +454,7 @@ struct NodeHeader {
 
 impl NodeHeader {
     fn size() -> Bytes {
-        Bytes::from(size_of::<Self>() as u64)
+        Bytes::from(core::mem::size_of::<Self>() as u64)
     }
 }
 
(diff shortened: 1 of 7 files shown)


dsarlis:
I think the CI recognizes only `perf` and not `opt` in the PR title.

Also, can you please un-commit the changes to `canbench_results.yml`? I'd like for the bot to first make the comment on what's the expected change (which also leaves a good paper trail of gains) and then if we agree this is a beneficial change, we can commit back the changes to results (not entirely sure the bot will post a comment if the results file is already changed).

src/btreemap/node.rs
```diff
@@ -474,14 +458,47 @@ impl NodeHeader {
     }
 }
 
-// The value in a K/V pair.
+/// The value in a K/V pair.
 #[derive(Debug)]
-enum Value {
-    // The value's encoded bytes.
-    ByVal(Vec<u8>),
+pub(crate) struct Value {
```
dsarlis, line 463: Just to confirm: do we have enough test coverage to catch cases where the invariant we list would be violated?

dsarlis:
> > can you please un-commit the changes to canbench_results.yml
> 
> Looks like the bot has added the comment and the comment says `canbench_results.yml is up to date` meaning the numbers that the bot calculated match the numbers that my local run produced. Nevertheless, I'm happy to uncommit and recommit the file if you think that makes sense.

No that's fine. I think this works even better then, you can already commit the change and we can still see the comment and validate it's a change we want.

src/btreemap/node.rs
frankdavid: The constructors or `Value` ensured that this was already the case, however, I moved back to an enum in order to encode it in the type system.

dsarlis (approved):

src/btreemap/node.rs
dsarlis: Even better to enforce it on the type system level, thanks!

PR #246, opened 2024-11-20
perf: Add new efficient API read_to_vec

frankdavid:
I found that a source of significant performance loss is the `read` method of `Memory`. The `read` method takes a mutable buffer which it fills with values read from the stable memory. According to Rust rules, the buffer passed to `read` must be initialized before it's passed to `read` (buffers containing uninitialized values are unsound and can cause UB).

The usual pattern is to create a properly sized `Vec`, eg. by using `vec![0; size]` or `vec.resize(size, 0)` and pass that to `read`. However, initializing the bytes with values that get overwritten by `read` is only necessary in order to be sound and requires significant number of instructions.

This PR introduces a new method `read_to_vec` which allows passing in a `Vec` with any size (most of the time an empty `Vec` will be used). Implementations can be more efficient by reading directly into the `Vec` and skipping initialization. This can lead to instruction reductions of up to 20%.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,6 +55,16 @@ pub trait Memory {
     /// and replaces the corresponding bytes in dst.
     fn read(&self, offset: u64, dst: &mut [u8]);
 
+    /// Copies len bytes of data starting from offset out of the stable memory into dst.
+    /// Implementations must not make any assumptions about dst (e.g. length, contents, capacity).
+    /// Callers are allowed to pass empty vectors. After the method returns, dst.len() == len.
+    /// This method is an alternative to read which does not require initializing a buffer and may
+    /// therefore be faster.
+    fn read_to_vec(&self, offset: u64, len: usize, dst: &mut std::vec::Vec<u8>) {
+        dst.resize(len, 0);
+        self.read(offset, &mut dst[..]);
+    }
+
     /// Copies the data referred to by src and replaces the
     /// corresponding segment starting at offset in the stable memory.
     fn write(&self, offset: u64, src: &[u8]);
--- a/canbench_results.yml
+++ b/canbench_results.yml
@@ -1,595 +1,595 @@
 benches:
   btreemap_get_blob_128_1024:
     total:
-      instructions: 879302205
+      instructions: 838312230
       heap_increase: 0
       stable_memory_increase: 0
     scopes: {}
   btreemap_get_blob_128_1024_v2:
     total:
-      instructions: 986234201
+      instructions: 925296023
       heap_increase: 0
       stable_memory_increase: 0
     scopes: {}
   btreemap_get_blob_16_1024:
     total:
-      instructions: 315520793
+      instructions: 312574719
       heap_increase: 0
       stable_memory_increase: 0
     scopes: {}
   btreemap_get_blob_16_1024_v2:
     total:
-      instructions: 424439334
+      instructions: 389362994
       heap_increase: 0
       stable_memory_increase: 0
     scopes: {}
   btreemap_get_blob_256_1024:
     total:
-      instructions: 1410478508
+      instructions: 1341541199
       heap_increase: 0
       stable_memory_increase: 0
     scopes: {}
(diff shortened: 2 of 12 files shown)


src/lib.rs
```diff
@@ -55,6 +55,16 @@ pub trait Memory {
...
 
+    /// Copies len bytes of data starting from offset out of the stable memory into dst.
+    /// Implementations must not make any assumptions about dst (e.g. length, contents, capacity).
+    /// Callers are allowed to pass empty vectors. After the method returns, dst.len() == len.
+    /// This method is an alternative to read which does not require initializing a buffer and may
+    /// therefore be faster.
+    fn read_to_vec(&self, offset: u64, len: usize, dst: &mut std::vec::Vec<u8>) {
+        dst.resize(len, 0);
+        self.read(offset, &mut dst[..]);
+    }
```
dsarlis, line 66: I am not sure if we want to have this part of `Memory` trait. It depends on how religious we want to be here.

The `Memory` trait is supposed to model the operations that are available for stable memory which in turn models the operations available for heap memory in Wasm. There's no `read_to_vec` so following that logic, maybe we avoid adding it to the trait.

We can probably have it as a free function or you think there's a significant advantage in having it on the trait?

frankdavid: The benefit of having it on the trait is that implementors of the trait can decide to override it in an efficient way. For example, `Ic0StableMemory` can read directly read into a `Vec` skipping initialization.

> is supposed to model the operations that are available for stable memory

That's a valid point. Another alternative could be changing the read method to read into  `dst: &mut [MaybeUninit<u8>]` which is technically the truth. It's unfortunately an unsolved problem in Rust to create uninitialized `[u8]` slices (something totally acceptable in C or C++). The documentation is very explicit about warning people not to create slices with uninitialized values even if those values will never be read.

frankdavid: I tested what if we replace read_to_vec with `unsafe fn read_unsafe(&self, offset: u64, dst: *mut u8, count: usize)`

It allows for even better optimizations than `read_to_vec` leading to up to 40% instruction reduction. This function is unsafe because clients must promise that `count` number of bytes after dst are writeable. 

(We could still keep `read_to_vec` but make it into an associated method like you mentioned. It would provide a safe way and efficient way for clients to use write.)

dsarlis: So, to make sure I understand your suggestion correctly, it would be:
1. Add `unsafe fn read_unsafe` for maximum performance if clients know what they are doing -- not super excited because I think it can be a big trap but as long as we provide safe alternatives I'm not strongly opposed.
2. Add `read_to_vec` as an associated method, not on the trait.
3. We would also leave the original `read` but it's the slowest of them all?

Thanks for all the exploration around possible options.

frankdavid: That's correct. I'd keep the original `read` because sometimes you have a fixed-size array that you want to fill. If the array is small, initializing it is fast and using a vector is unnecessary (and slower). `read` is the best option for this use-case. 

When reading a larger blob, it's better not to allocate on the stack, so reading into a `Vec` is the best approach. `read_to_vec` can be used for that.

And `Memory` implementations must be able to read into uninitialized buffers (pointer + length) which is trivial for the mostly used Ic0Memory implementation. The default implementation will just initialize the buffer with 0-s first and then call the existing `read()`. If we make `read_to_vec` into an associated function, we don't have to pollute the Memory interface with `Vec`.

>  if clients know what they are doing

The API will be the same as for example `std::ptr::copy` with the same requirements. And we can provide safe wrappers while staying efficient.

I'll send a new PR with this new approach.

rikonor: Bit of a random question, but we have several canisters that use stable-structures, and we'd love to benefit from these improvements - would we be able to just by upgrading the crate once this change is out?

dsarlis: @rikonor Yeah, we'll need to cut a new release because there's a few optimizations already included that are unreleased. Maybe we can target to make a release after this change with `read_to_vec` is merged.

PR #248, opened 2024-11-21
perf: Add new efficient APIs read_unsafe and read_to_vec

frankdavid:
I found that a source of significant performance loss is the read method of `Memory`. The `read` method takes a mutable buffer which it fills with values read from the stable memory. According to Rust rules, the buffer passed to read must be initialized before it's passed to read (buffers containing uninitialized values are unsound and can cause UB).

The usual pattern is to create a properly sized Vec, eg. by using `vec![0; size]` or `vec.resize(size, 0)` and pass that to `read`. However, initializing the bytes with values that get overwritten by `read` is only necessary in order to be sound and requires significant number of instructions.

This PR introduces a new method `read_unsafe` which allows passing in a raw pointer and a `count` parameter. Implementations can be more efficient by reading directly and skipping initialization. This can lead to instruction reductions of up to 40%.

The PR also introduces a helper method `read_to_vec` which is a safe wrapper around `read_unsafe` for the most common use-case: reading into a `Vec`. Clients can for example pass an empty `Vec` and profit from the extra efficiency without having to call unsafe methods.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -83,7 +83,7 @@ pub trait Memory {
 pub trait MemoryExt {
     /// Copies `count` bytes of data starting from offset out of the stable memory into `dst`.
     ///
-    /// Callers are allowed to pass vectors in state (e.g. empty vectors).
+    /// Callers are allowed to pass vectors in any state (e.g. empty vectors).
     /// After the method returns, `dst.len() == count`.
     /// This method is an alternative to `read` which does not require initializing a buffer and may
     /// therefore be faster.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -92,7 +92,7 @@ pub trait Memory {
 #[inline]
 fn read_to_vec<M: Memory>(m: &M, addr: Address, dst: &mut std::vec::Vec<u8>, count: usize) {
     dst.clear();
-    dst.reserve(count);
+    dst.reserve_exact(count);
     unsafe {
         m.read_unsafe(addr.get(), dst.as_mut_ptr(), count);
         // SAFETY: read_unsafe guarantees to initialize the first `count` bytes
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,6 +26,7 @@ pub use file_mem::FileMemory;
 pub use ic0_memory::Ic0StableMemory;
 use std::error;
 use std::fmt::{Display, Formatter};
+use std::mem::MaybeUninit;
 pub use storable::Storable;
 use types::Address;
 pub use vec_mem::VectorMemory;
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -46,20 +46,64 @@ pub trait Memory {
...
+    /// Callers are allowed to pass vectors in any state (e.g. empty vectors).
+    /// After the method returns, `dst.len() == count`.
+    /// This method is an alternative to `read` which does not require initializing a buffer and may
+    /// therefore be faster.
+    fn read_to_vec(&self, offset: u64, dst: &mut std::vec::Vec<u8>, count: usize);
+}
+
+impl<M: Memory> MemoryExt for M {
+    #[inline]
+    fn read_to_vec(&self, offset: u64, dst: &mut std::vec::Vec<u8>, count: usize) {
```
berestovskyy, line 97: nit: WDYT about just `read_to_vec(offset, count) -> Vec<u8>`? Then we don't need to clear the vector...

frankdavid: Passing an existing `Vec` allows clients to reuse existing allocations which results in better performance.

berestovskyy: Makes sense. Although similar standard library functions either [append the data](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_to_string) or [return a new object](https://doc.rust-lang.org/std/io/fn.read_to_string.html) (with the same performance note you mentioned)...

berestovskyy: Ah I see that `read_entry_to` has the similar replacing semantics. Ok then...

berestovskyy (commented):
Thanks, LGTM modulo a few comments. Should we also consider adding some tests for `read_to_vec`?

src/lib.rs
```diff
@@ -46,20 +46,64 @@ pub trait Memory {
...
+        dst.clear();
+        dst.reserve(count);
+        unsafe {
+            self.read_unsafe(offset, dst.as_mut_ptr(), count);
+            dst.set_len(count);
```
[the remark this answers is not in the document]
frankdavid: From `set_len` documentation: `Safety: [...] The elements at old_len..new_len must be initialized.`
Since `reserve` does not initialize, this would be a violation of the safety rules. If this was allowed, we wouldn't need the new `read_unsafe` method in the first place, we could always just pass an uninitialized slice to read (like we can do in C or C++). It's unfortunately illegal in Rust.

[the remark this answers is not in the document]
berestovskyy: Yeah, it's an undefined behavior even for `u8` in Rust...

berestovskyy (approved):

berestovskyy (approved):

PR #249, opened 2024-11-22
perf: Add `into_bytes` to `Storable` to avoid cloning where possible

hpeebles:
In OpenChat I have just introduced a `StableBTreeMap` where the values are `Vec<u8>`.
I saw that when calling `map.insert(..)` the bytes were being cloned because `Storable::to_bytes` take a reference to `self` so for `Vec<u8>` the response is `Cow::Borrowed(self)`, then when calling `into_owned` the bytes are cloned.

This PR adds an optional `into_bytes` method to the `Storable` trait which by default simply calls into `to_bytes`.
Implementing `into_bytes` allows for values to be inserted with cloning the bytes.

The downside is that I've had to add the `Sized` trait to `Storable` which is a breaking change and may not be desired.

### DIFF
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -353,7 +353,7 @@ where
     ///   key.to_bytes().len() <= max_size(Key)
     ///   value.to_bytes().len() <= max_size(Value)
     pub fn insert(&mut self, key: K, value: V) -> Option<V> {
-        let value = value.to_bytes_checked().into_owned();
+        let value = value.into_bytes_checked();
 
         let root = if self.root_addr == NULL {
             // No root present. Allocate one.
--- a/src/storable.rs
+++ b/src/storable.rs
@@ -10,7 +10,7 @@ mod tuples;
 mod tests;
 
 /// A trait with convenience methods for storing an element into a stable structure.
-pub trait Storable {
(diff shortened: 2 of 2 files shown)


hpeebles:
This same idea may also be of use within `ic_stable_structures::vec::Vec` where `push` takes a reference which means there will almost always be cloning.

hpeebles:
The same idea may apply to `ic_stable_structures::vec::Vec` where `push` takes a reference so there will (almost) always be cloning.

dsarlis:
I'm really not sure about the `Size` requirement. I feel like it might not be worth it. Is there no other way to go around what happens for `Vec<u8>`? Maybe we introduce some shortcut of the form "if the value to be encoded is a `Vec<u8>` then skip serialization altogether" since storing the raw bytes should be fine.

dsarlis:
Also, using `Cow` to begin with was to avoid unnecessary cloning until the point where you actually need it. So you're saying you don't want to have any cloning at all? Typically (in most Rust libraries I am aware of), during serialization you don't give up ownership, you pass a reference and that means there needs to be some cloning at some point.

hpeebles:
Woah I didn't get those canbench regressions when running locally!
Let me have a look at what happened here.

PR #250, opened 2024-11-24
fix: Fix Vec Iter::count() after next_back()

frankdavid:
There is a bug in the current implementation that returns the count until the end of the vector even if some elements have been taken from the iterator using `next_back()` (see the testcase in the PR).

### DIFF
--- a/src/vec/tests.rs
+++ b/src/vec/tests.rs
@@ -227,6 +227,13 @@ fn test_iter() {
     assert_eq!(sv.iter().skip(3).count(), 0);
     assert_eq!(sv.iter().skip(4).count(), 0);
     assert_eq!(sv.iter().skip(usize::MAX).count(), 0);
+
+    {
+        assert_eq!(sv.len(), 3);
+        let mut iter = sv.iter();
+        iter.next_back();
+        assert_eq!(iter.count(), 2);
+    }
 }
 
 // A struct with a bugg implementation of storable where the max_size can
--- a/src/vec/tests.rs
+++ b/src/vec/tests.rs
@@ -227,11 +227,25 @@ fn test_iter() {
     assert_eq!(sv.iter().skip(3).count(), 0);
     assert_eq!(sv.iter().skip(4).count(), 0);
     assert_eq!(sv.iter().skip(usize::MAX).count(), 0);
+}
 
+#[test]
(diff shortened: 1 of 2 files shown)


src/vec/tests.rs
```diff
@@ -227,6 +227,13 @@ fn test_iter() {
     assert_eq!(sv.iter().skip(3).count(), 0);
     assert_eq!(sv.iter().skip(4).count(), 0);
     assert_eq!(sv.iter().skip(usize::MAX).count(), 0);
+
+    {
+        assert_eq!(sv.len(), 3);
+        let mut iter = sv.iter();
+        iter.next_back();
+        assert_eq!(iter.by_ref().count(), 2);
```
frankdavid, line 235: Before this PR, this returns 3. The stdlib `Vec` returns 2 so we should also return 2.

src/vec/tests.rs
```diff
@@ -227,6 +227,13 @@ fn test_iter() {
     assert_eq!(sv.iter().skip(3).count(), 0);
     assert_eq!(sv.iter().skip(4).count(), 0);
     assert_eq!(sv.iter().skip(usize::MAX).count(), 0);
+
+    {
+        assert_eq!(sv.len(), 3);
+        let mut iter = sv.iter();
+        iter.next_back();
+        assert_eq!(iter.count(), 2);
```
frankdavid, line 235: Before this PR, this returns 3. The stdlib `Vec` returns 2 so we should also return 2.

src/vec/tests.rs
```diff
@@ -229,6 +229,27 @@ fn test_iter() {
...
+fn test_iter_count() {
+    let sv = StableVec::<u64, M>::new(M::default()).unwrap();
+    sv.push(&1).unwrap();
+    sv.push(&2).unwrap();
+    sv.push(&3).unwrap();
+    sv.push(&4).unwrap();
+    {
+        let mut iter = sv.iter();
+        iter.next_back();
+        assert_eq!(iter.count(), 3);
```
frankdavid, line 242: Before this PR, this returns 4. The stdlib `Vec` returns 3 so we should also return 3.

berestovskyy (approved):
Thanks, great catch!

PR #251, opened 2024-11-25
perf: Entry API POC

frankdavid:
This is a proposal for preferring iterating over Entries rather than `(K, V)` tuples. As the benchmarks show, there is a considerable speed gain in scans, i.e. if some code only needs to access certain values depending on the key. If values are large / deserialization of values is slow, the speed gain can be even higher.

However, this is a **breaking change.** 

Before:
```
    for (k, v) in map.range(..) {
        // Do something with k and v
    }
```

After:
```
    for entry in map.range(..) {
        // Do something with entry.key() and entry.value()
    }
```

Affected functions: `range()`, `iter()`, `iter_upper_bound()`

The backwards-compatible alternative would be introducing the API under new method names, however that would convolute the API surface.

This PR is not complete yet, tests have been disabled for the time of the discussion.


berestovskyy:
Should we maintain backward compatibility by adding a new function like `entry_range` to create an iterator over the range or entries?

frankdavid:
I'm concerned that it'd make the API quite convoluted: 
`iter()` would need `iter_entries()` or `entry_iter()`
`range()` would need `entry_range()`
`iter_upper_bound()` would need `entry_iter_upper_bound()`

Every time we add a new function which returns an iterator, we'd need to duplicate it (not even mentioning variants corresponding to `keys()` and `values()`).

I was considering a single iterator API with iterator transformers, like
`iter().as_key_value_tuples()` or `range(...).as_key_value_tuples()` (bad naming, only for illustration)

berestovskyy:
Hey David, we've discussed this internally, and we're in favor of adopting this approach, as it clearly brings a significant performance improvement. Do you think it makes sense to discuss the new API with a wider group of developers?

PR #254, opened 2024-11-29
chore: Bump Rust version to 1.78 in rust-toolchain

frankdavid:
A transitive dependency (cargo-platform) has been updated and now requires min Rust version 1.78.

Our library can still be used with a lower Rust version but our tests/benchmarks will be run with 1.78.

### DIFF
--- a/canbench_results.yml
+++ b/canbench_results.yml
@@ -1,595 +1,595 @@
 benches:
   btreemap_get_blob_128_1024:
     total:
-      instructions: 816917041
+      instructions: 821164751
       heap_increase: 0
       stable_memory_increase: 0
     scopes: {}
   btreemap_get_blob_128_1024_v2:
     total:
-      instructions: 891561192
+      instructions: 899572730
       heap_increase: 0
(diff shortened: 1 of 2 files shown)


berestovskyy (approved):
Approving to unblock you. We should figure out why, for example, `vec_insert_u64` regresses ~5%...

frankdavid:
> We should figure out why, for example, vec_insert_u64 regresses ~5%...

I think there are two things at play here:
- Rust optimizes overall performance, individual benchmarks may get slower or faster depending on code generation.
- This effect is amplified even more in our case because we have our own way of measuring performance via instruction count and custom weighting which sometimes doesn't correspond to real world performance (for example, it doesn't take into account CPU caches and branch prediction, furthermore function call overhead is discounted). As I was exploring places in the code that could be optimized, it happened that a change that I expected to greatly improve performance, actually increased the instruction count. I can imagine that there are changes in the Rust stdlib which improved the metrics that the stdlib maintainers optimize for but made things worst according to our metrics.

PR #256, opened 2024-12-02
perf: Support read_unsafe in VirtualMemory and RestrictedMemory

frankdavid:
The `VirtualMemory` implementation didn't make use of the newly introduced read_unsafe until now. Since many clients use stable structures via the `VirtualMemory`, we can speed those up by using `read_unsafe` in `VirtualMemory`.


berestovskyy (approved):
Looks good, thanks!

PR #257, opened 2024-12-02
chore: Format canbench output so diffs can be seen more easily

frankdavid:
Since Github does not support colored text, I chose to use colored emojis.

Demo:
Screenshot 2024-12-02 at 17 25 12


frankdavid:
Demo:
Screenshot 2024-12-02 at 17 25 12

berestovskyy (approved):
Nice, thanks!

PR #260, opened 2024-12-04
perf: Add bucket cache in MemoryManager

frankdavid:
Since buckets are quite big (minimum 64kb and 8MB by default), subsequent accesses will likely touch the same bucket. It's worth caching the last accessed bucket.


berestovskyy (approved):
Looks good, thanks!

PR #276, opened 2025-04-08
chore: cleanup benchmarks

maksymar:
This PR improves readability of similar benchmarks by using macro rule.

### DIFF
--- a/benchmarks/src/btreemap.rs
+++ b/benchmarks/src/btreemap.rs
@@ -72,9 +72,74 @@ bench_tests! {
 
     btreemap_insert_u64_blob_8,        insert_helper_v1, u64,   Blob8;
     btreemap_insert_u64_blob_8_v2,     insert_helper_v2, u64,   Blob8;
-
     btreemap_insert_blob_8_u64,        insert_helper_v1, Blob8,   u64;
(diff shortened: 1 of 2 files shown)


benchmarks/src/btreemap.rs
```diff
@@ -5,212 +5,141 @@ use ic_stable_structures::{storable::Blob, BTreeMap, DefaultMemoryImpl, Memory,
...
+type Blob16 = Blob<16>;
+type Blob32 = Blob<32>;
+type Blob64 = Blob<64>;
+type Blob128 = Blob<128>;
+type Blob256 = Blob<256>;
+type Blob512 = Blob<512>;
+type Blob1024 = Blob<1024>;
+
+// Benchmarks inserting blobs into a BTreeMap.
+bench_tests! {
```
berestovskyy, line 31: nit: the usual issue with the macros is that the text inside them is not auto-formatted. Below, the shorter `Blobs` seem to be not aligned with the longer blobs.

berestovskyy (approved):
Looks good, thanks!

benchmarks/src/btreemap.rs
maksymar: I did that intentionally, somehow this is easier for me to read.

PR #279, opened 2025-04-08
Draft: POC [EXC-2017] Cache the first and the last key-value pair in StableBTreeMap

dragoljub-duric:
This PR adds caching for the first and last key-value pairs, which should speed up `fn first_key_value` and `fn last_key_value`. The only downside is that these functions will now require a mutable reference on `StableBTreeMap`, since caching can mutate `StableBTreeMap`.

Canbench results:
---------------------------------------------------
---------------------------------------------------

Benchmark: btreemap_first_key_value
  total:
    instructions: 4.98 M (improved by 95.75%)
    heap_increase: 0 pages (no change)
    stable_memory_increase: 0 pages (no change)

---------------------------------------------------

Benchmark: btreemap_first_key_value_pop_first
  total:
    instructions: 266.67 M (improved by 48.62%)
    heap_increase: 0 pages (no change)
    stable_memory_increase: 0 pages (no change)

---------------------------------------------------

Benchmark: btreemap_first_key_value_pop_last
  total:
    instructions: 262.13 M (improved by 45.75%)
    heap_increase: 0 pages (no change)
    stable_memory_increase: 0 pages (no change)

---------------------------------------------------

Benchmark: btreemap_first_key_value_pop_first_blob1024
  total:
    instructions: 285.63 M (improved by 55.91%)
    heap_increase: 0 pages (no change)
    stable_memory_increase: 0 pages (no change)

---------------------------------------------------

Benchmark: btreemap_first_key_value_pop_first_once_blob1024
  total:
    instructions: 514.56 M (0.93%) (change within noise threshold)
    heap_increase: 0 pages (no change)
    stable_memory_increase: 10 pages (no change)

---------------------------------------------------

Benchmark: btreemap_last_key_value
  total:
    instructions: 9.94 M (improved by 96.74%)
    heap_increase: 0 pages (no change)
    stable_memory_increase: 0 pages (no change)

---------------------------------------------------

Benchmark: btreemap_last_key_value_pop_first
  total:
    instructions: 509.65 M (improved by 47.60%)
    heap_increase: 0 pages (no change)
    stable_memory_increase: 0 pages (no change)

---------------------------------------------------

Benchmark: btreemap_last_key_value_pop_last
  total:
    instructions: 500.35 M (improved by 48.16%)
    heap_increase: 0 pages (no change)
    stable_memory_increase: 0 pages (no change)

---------------------------------------------------

Benchmark: btreemap_last_key_value_pop_last_blob1024
  total:
    instructions: 280.63 M (improved by 57.29%)
    heap_increase: 0 pages (no change)
    stable_memory_increase: 0 pages (no change)

---------------------------------------------------

Benchmark: btreemap_last_key_value_pop_last_once_blob1024
  total:
    instructions: 483.40 M (0.99%) (change within noise threshold)
    heap_increase: 0 pages (no change)
    stable_memory_increase: 10 pages (no change)

---------------------------------------------------

### DIFF
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -367,13 +367,13 @@ where
     pub fn insert(&mut self, key: K, value: V) -> Option<V> {
         if let Some((first_key, _)) = &self.first_key_value {
             if key <= *first_key {
-                self.first_key_value = None;
+                self.first_key_value = Some((key.clone(), value.to_bytes_checked().into_owned()));
             }
         }
 
         if let Some((last_key, _)) = &self.last_key_value {
             if key >= *last_key {
-                self.last_key_value = None;
+                self.last_key_value = Some((key.clone(), value.to_bytes_checked().into_owned()));
             }
         }
 
@@ -608,7 +608,7 @@ where
             return Some((k.clone(), V::from_bytes(Cow::Owned(encoded_v.clone()))));
         }
 
-        let root = self.load_node(self.root_addr);
+        let root: Node<K> = self.load_node(self.root_addr);
         let (k, encoded_v) = root.get_min(self.memory());
         self.first_key_value = Some((k.clone(), encoded_v.clone()));
         Some((k, V::from_bytes(Cow::Owned(encoded_v))))
@@ -645,18 +645,6 @@ where
             return None;
         }
 
-        if let Some((first_key, _)) = &self.first_key_value {
-            if first_key == key {
-                self.first_key_value = None;
-            }
-        }
-
-        if let Some((last_key, _)) = &self.last_key_value {
-            if last_key == key {
-                self.last_key_value = None;
-            }
-        }
-
         let root_node = self.load_node(self.root_addr);
         self.remove_helper(root_node, key)
             .map(Cow::Owned)
@@ -669,14 +657,8 @@ where
             return None;
         }
 
-        if self.length == 1 {
-            self.first_key_value = None;
-        }
-
-        self.last_key_value = None;
-
         let root = self.load_node(self.root_addr);
-        let (max_key, _) = root.get_max(self.memory());
+        let (max_key, _) = self.last_key_value().unwrap();
         self.remove_helper(root, &max_key)
             .map(|v| (max_key, V::from_bytes(Cow::Owned(v))))
     }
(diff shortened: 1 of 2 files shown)


src/btreemap.rs
```diff
@@ -575,23 +599,35 @@ where
 
     /// Returns the first key-value pair in the map. The key in this
     /// pair is the minimum key in the map.
-    pub fn first_key_value(&self) -> Option<(K, V)> {
+    pub fn first_key_value(&mut self) -> Option<(K, V)> {
```
maksymar, line 602: caching does not have to change the function API. the general approach in rust for caching is to use internal mutability with RefCell<...>.

src/btreemap.rs
```diff
@@ -353,6 +365,18 @@ where
     ///   key.to_bytes().len() <= max_size(Key)
     ///   value.to_bytes().len() <= max_size(Value)
     pub fn insert(&mut self, key: K, value: V) -> Option<V> {
+        if let Some((first_key, _)) = &self.first_key_value {
```
maksymar, line 368: Since both first and last entry caches are touched in this method I suggest you create a single `first_last_entry_cache` wrapped by `RefCell` so that it calls `borrow_mut()` only once and then decide which part to update. inside you can store a pair and handle a special case when BTM.len()==1 to always use one of the cells.

src/btreemap.rs
```diff
@@ -575,23 +599,35 @@ where
...
     /// Returns the last key-value pair in the map. The key in this
     /// pair is the maximum key in the map.
-    pub fn last_key_value(&self) -> Option<(K, V)> {
+    pub fn last_key_value(&mut self) -> Option<(K, V)> {
         if self.root_addr == NULL {
             return None;
         }
+
+        if let Some((k, encoded_v)) = &self.last_key_value {
+            return Some((k.clone(), V::from_bytes(Cow::Owned(encoded_v.clone()))));
```
maksymar, line 625: would it be possible to store the actual result of this not doing the work every time?

src/btreemap.rs
```diff
@@ -3071,4 +3120,137 @@ mod test {
...
     }
+
+    #[test]
+    fn first_key_value_correctness() {
+        btree_test(|mut btree| {
+            // Initially, the tree is empty.
+            assert_eq!(btree.first_key_value(), None);
+
+            // Insert some elements.
+            assert_eq!(btree.insert(b(&[3]), b(&[30])), None);
```
maksymar, line 3131: It'd be better to write a few loops to check the invariants of each case (insert + read, remove + read, pop + read), eg

```rs
let n = 100;
let key = |i| b(&[i]);
let value = |i| b(&[i * 10]);

// insert + read
for i in 0..n {
  assert_eq!(btree.insert(key(i), value(i)), None);
  assert_eq!(btree.first_key_value(), Some(key(0)));
  assert_eq!(btree.last_key_value(), Some(key(i)));
}

// remove + read
for i in 0..n {
  assert_eq!(btree.remove(key(i), value(i)), Some(value(i)));
  assert_eq!(btree.first_key_value(), Some(key(i+1)));
}

...
```

PR #283, opened 2025-04-16
test: add multi‑type key/value support to BTreeMap tests

maksymar:
Run each `BTreeMap` test for multiple key/value type pairs (bounded and unbounded) by using `btree_test!` macro.

No functional changes, only test code is modified.

Changes:
- add `Build` trait to generate key/value types from `u32` value in monotonic way
- add `btree_test!` macro to run each test on a set of key/value pairs (bounded/unbounded, set-like, map-like)
- add unified key/value helpers to abstract the actual type and improve readability of each test by using `key(i)`/`value(i)` instead of `b(&[1,2,3])`

Before:
```rs
assert_eq!(btree.insert(b(&[1,2,3]), b(&[4,5,6])), None);
assert_eq!(btree.get(&b(&[1,2,3])), Some(b(&[4,5,6])));
assert_eq!(
    btree.range(b(&[1, 5])..=b(&[2, 6])).collect::<Vec<_>>(),
    vec![
        (b(&[1, 6]), b(&[])),
        (b(&[1, 8]), b(&[])),
        (b(&[1, 10]), b(&[])),
        (b(&[2, 1]), b(&[])),
        (b(&[2, 2]), b(&[])),
        (b(&[2, 3]), b(&[])),
        (b(&[2, 4]), b(&[])),
        (b(&[2, 5]), b(&[])),
        (b(&[2, 6]), b(&[])),
    ]
);
```

After:
```rs
let (key, value) = (|i| K::build(i), |i| V::build(i));

assert_eq!(btree.insert(key(123), value(456)), None);
assert_eq!(btree.get(&key(123)), Some(value(456)));
assert_eq!(
    collect(btree.range(key(15)..=key(26))),
    vec![
        (key(16), V::empty()),
        (key(18), V::empty()),
        (key(19), V::empty()),
        (key(21), V::empty()),
        (key(22), V::empty()),
        (key(23), V::empty()),
        (key(24), V::empty()),
        (key(25), V::empty()),
        (key(26), V::empty()),
    ]
);
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,3 @@
 .vscode/
 .idea/
 **/*~
-
-tmp/
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@
 .vscode/
 .idea/
 **/*~
+
+tmp/
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -1236,16 +1236,6 @@ mod test {
         }
     }
 
-    type MonotonicString32 = String;
-    impl Builder for MonotonicString32 {
-        fn build(i: u32) -> Self {
-            format!("{i:0>32}")
-        }
-        fn empty() -> Self {
-            String::new()
-        }
-    }
-
     type MonotonicVec32 = Vec<u8>;
     impl Builder for MonotonicVec32 {
         fn build(i: u32) -> Self {
@@ -1253,6 +1243,16 @@ mod test {
         }
         fn empty() -> Self {
             Vec::new()
+        }
+    }
+
+    type MonotonicString32 = String;
+    impl Builder for MonotonicString32 {
+        fn build(i: u32) -> Self {
+            format!("{i:0>32}")
+        }
+        fn empty() -> Self {
+            String::new()
         }
     }
 
@@ -1324,14 +1324,14 @@ mod test {
                 // Set.
                 verify_and_run!($runner, u32, ());
                 verify_and_run!($runner, Blob<10>, ());
-                verify_and_run!($runner, MonotonicString32, ());
                 verify_and_run!($runner, MonotonicVec32, ());
+                verify_and_run!($runner, MonotonicString32, ());
 
                 // Map.
                 verify_and_run!($runner, u32, Blob<20>);
                 verify_and_run!($runner, Blob<10>, Blob<20>);
-                verify_and_run!($runner, MonotonicString32, Blob<20>);
                 verify_and_run!($runner, MonotonicVec32, Blob<20>);
+                verify_and_run!($runner, MonotonicString32, Blob<20>);
             }
         };
     }
--- a/src/btreemap.rs
+++ b/src/btreemap.rs
@@ -1160,6 +1160,7 @@ mod test {
         storable::{Blob, Bound as StorableBound, Storable},
         VectorMemory,
     };
+    use core::panic;
     use std::borrow::Cow;
     use std::cell::RefCell;
     use std::convert::TryFrom;
@@ -1196,13 +1197,7 @@ mod test {
         }
     }
 
-    /// Creates a key from a u32.
-    fn k<T: Make>(i: u32) -> T {
-        T::make(i)
-    }
-
-    /// Creates a value from a u32.
-    fn v<T: Make>(i: u32) -> T {
+    fn make<T: Make>(i: u32) -> T {
         T::make(i)
     }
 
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -6,3 +6,5 @@
 .vscode/
 .idea/
 **/*~
+
+tmp/
```
maksymar, line 10: I often have some temporary files from other branches experiments that I don't want to lose and don't want to commit to the current branch. I suppose it's ok to have a `tmp/` in gitignore in genera.

src/btreemap.rs
```diff
@@ -1161,156 +1161,317 @@ mod test {
...
+        ($runner:ident, $K:ty, $V:ty) => {{
+            verify_monotonic::<$K>();
+            $runner::<$K, $V>();
+        }};
+    }
 
-            // Reload the btree
-            let btree = BTreeMap::init(btree.into_memory());
+    /// Macro to apply a test function to a predefined grid of key/value types.
+    macro_rules! btree_test {
```
maksymar, line 1308: FYI: This macro adds key/value type dimension to every test.

src/btreemap.rs
```diff
@@ -1161,156 +1161,317 @@ mod test {
...
+            assert_eq!(
+                output, expected,
+                "monotonic_buffer::<4>({}) returned {:?}, expected {:?}",
+                input, output, expected
+            );
+        }
+    }
+
+    /// A trait to construct a value from a u32.
+    trait Builder {
```
maksymar, line 1199: FYI: this trait provides a type abstraction and improves readability of each test.

In order not to write `K::build(i)`/`V::build(i)` I prefer to add a `key`/`value` closures to each test to write `key(i)`/`value(i)`.

src/btreemap.rs
```diff
@@ -1161,156 +1161,317 @@ mod test {
...
-    #[test]
-    fn allocations() {
-        btree_test(|mut btree| {
+    fn allocations_without_split<K: TestKey, V: TestValue>() {
+        let key = |i| K::build(i);
+        run_btree_test(|mut btree| {
+            assert_eq!(btree.allocator.num_allocated_chunks(), 0);
+
+            assert_eq!(btree.insert(key(1), V::empty()), None);
+            assert_eq!(btree.allocator.num_allocated_chunks(), 1);
```
maksymar, line 1460: FYI: here the size of value object may affect `num_allocated_chunks` therefore it has to be minimal (empty) size.

src/btreemap.rs
```diff
@@ -1161,156 +1161,329 @@ mod test {
...
     use std::rc::Rc;
 
-    pub(crate) fn make_memory() -> Rc<RefCell<Vec<u8>>> {
-        Rc::new(RefCell::new(Vec::new()))
+    /// A helper function to clone and collect borrowed key/value pairs into a `Vec`.
+    fn collect_kv<'a, K: Clone + 'a, V: Clone + 'a>(
+        iter: impl Iterator<Item = (&'a K, &'a V)>,
+    ) -> Vec<(K, V)> {
+        iter.map(|(k, v)| (k.clone(), v.clone())).collect()
+    }
```
michael-weigelt, line 1172: Is there a reason why the following would not work? 
```
x.cloned().collect()
```
instead of 
```
x.collect_kv()
```

src/btreemap.rs
```diff
@@ -1161,156 +1161,329 @@ mod test {
...
+
+    /// A helper function to collect owned key/value pairs into a `Vec`.
+    fn collect<K: Clone, V: Clone>(it: impl Iterator<Item = (K, V)>) -> Vec<(K, V)> {
+        it.collect()
+    }
```
michael-weigelt, line 1177: Is this just a shorthand such that you don't have to provide the type hint for `collect` at every call site?

src/btreemap.rs
maksymar: it's a shorthand to improve the readability of the assert, otherwise too much boilerplate code.

src/btreemap.rs
```diff
@@ -1161,156 +1161,329 @@ mod test {
...
+    // Define a trait for keys that need the full set of bounds.
+    trait TestKey: Storable + Ord + Clone + Builder + std::fmt::Debug {}
+    impl<T> TestKey for T where T: Storable + Ord + Clone + Builder + std::fmt::Debug {}
+
+    // Define a trait for values that need the full set of bounds.
+    trait TestValue: Storable + Clone + Builder + std::fmt::Debug + PartialEq {}
+    impl<T> TestValue for T where T: Storable + Clone + Builder + std::fmt::Debug + PartialEq {}
+
+    fn insert_get_init_preserves_data<K: TestKey, V: TestValue>() {
+        let (key, value) = (|i| K::build(i), |i| V::build(i));
```
michael-weigelt, line 1348:
```suggestion
        let (key, value) = (K::build, V::build);
```

michael-weigelt: also elsewhere

src/btreemap.rs
```diff
@@ -1321,96 +1494,47 @@ mod test {
...
-
-    #[test]
-    fn insert_split_node() {
-        btree_test(|mut btree| {
+    fn insert_split_node<K: TestKey, V: TestValue>() {
+        let (key, value) = (|i| K::build(i), |i| V::build(i));
+        run_btree_test(|mut btree| {
             for i in 1..=11 {
-                assert_eq!(btree.insert(b(&[i]), b(&[])), None);
+                assert_eq!(btree.insert(key(i), value(10)), None);
```
michael-weigelt, line 1509: Why value(10) and not empty()?

michael-weigelt (approved):

src/btreemap.rs
maksymar: oh, that's neat, thank you!

src/btreemap.rs
maksymar: I thought it brings more coverage when values are also different (you can imagine a bug where all the keys are pointing to the same value), but at some point there were so many changes that I just gave up updating every value.

michael-weigelt: there are some more cases where `b(&[])` becomes `value(0)`, not `empty()`, although that might be correct, idk.

PR #284, opened 2025-04-18
docs: improved documentation for BTreeMap

ielashi:


src/btreemap.rs
```diff
@@ -81,10 +81,149 @@ const DEFAULT_PAGE_SIZE: u32 = 1024;
...
+///
+/// **Important**: Each stable structure requires its own designated memory region. Attempting to
+/// initialize multiple structures with the same memory will lead to data corruption.
+///
+/// ### What NOT to do:
+///
+/// ```rust,no_run
+/// use ic_stable_structures::{BTreeMap, DefaultMemoryImpl};
+/// 
+/// // ERROR: Using the same memory for multiple BTreeMaps will corrupt data
```
berestovskyy, line 123: Seems like this points out that the `DefaultMemoryImpl::default()` should internally do `memory_manager.get(MemoryId::new(next))`. This is not related to this PR, but just to confirm WDYT @ielashi ?

berestovskyy (approved):

src/btreemap.rs
ielashi: Thanks for the suggestion @berestovskyy. I agree we should definitely think of ways of not requiring developers to set memory IDs manually. It's unfortunately not so straightforward. Where would you store this `next`? And how would you guarantee that these memory IDs are always the same? What if I add/remove structures between upgrades? We'd need to devise a mechanism that works across all these cases.

berestovskyy: I agree, it's not an easy change and should not be done in this PR. My high-level point is that to make the stable structures more user-friendly, we should avoid any default behaviors which might corrupt the data.

In this particular case, we first recommend the `DefaultMemoryImpl` for "almost all use cases", but a few lines later, we reveal that in fact for more than one stable structure we should use the `MemoryManager`.

ielashi: I see. The `MemoryManager` isn't a substitute for `DefaultMemoryImpl`. We should indeed use the `DefaultMemoryImpl` basically all the time, but it should be used _along with_ the MemoryManager. Maybe I can make this more clear in the docs.

PR #289, opened 2025-04-28
docs: add skeleton for stable structures book

ielashi:
There are a number of concepts in stable structures that need more thorough documentation, and I think it warrants some kind of "book" or reference guide that can be used by developers.

This commit adds the skeleton of what would become this book, and in the future we'll use mdbook to serve this documentation.

An example of what this book will look like:
https://proptest-rs.github.io/proptest/


docs/src/design-principles.md
```diff
@@ -0,0 +1,21 @@
...
+The library is built on several key principles:
+
+* **Radical simplicity**: Each data structure follows the most straightforward design that solves the problem at hand.
+This makes the code easier to understand, debug, and maintain.
+
+* **Backward compatibility**: Upgrading the library version must preserve the data.
+All data structures have a metadata section with the layout version, ensuring that new versions can read data written by old versions.
+
+* **No `pre_upgrade` hooks**: A bug in the `pre_upgrade` hook can make your canister non-upgradable.
+The best way to avoid this issue is not to have a `pre_upgrade` hook at all.
```
dsarlis, line 12: Should there be a link to the interface spec where the `pre_upgrade` hook is explained as this is the first mention of it in this book?

ielashi: Agreed. Done.

PR #290, opened 2025-04-28
docs: add memory trait and memory manger documentation to book

ielashi:
This commit adds a "Concepts" section to the book and introduces the `Memory` trait as well as the `MemoryManager`.

### DIFF
--- a/docs/src/concepts.md
+++ b/docs/src/concepts.md
@@ -1,0 +1,3 @@
+# Concepts
+
+This section covers fundamental concepts for understanding how stable structures work and how they can be used effectively.
--- a/docs/src/memory-manager.md
+++ b/docs/src/memory-manager.md
@@ -1,0 +1,37 @@
+# Memory Manager
+
+As mentioned in the previous section, each stable structure requires its own dedicated `Memory` instance.
+This is an intentional design decision that limits [the blast radius](./design-principles.md) of potential bugs, ensuring that issues only affect the specific data structure and its associated memory, not other structures.
+
+## Overview
+
+The Memory Manager enables the creation of up to 255 virtual memories from a single underlying memory instance.
+When used with stable memory, this allows you to maintain up to 255 separate stable structures, each with its own isolated memory space.
+
+## Usage
+
+The following example demonstrates how to use the Memory Manager to create multiple stable structures:
+
+```rust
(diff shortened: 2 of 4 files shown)


docs/src/concepts.md
```diff
@@ -0,0 +1,3 @@
+# Concepts
+
+This section covers fundamental concepts for understanding how stable structures work and how they can be used effectively.
```
dsarlis, line 3:
```suggestion
This section covers fundamental concepts for understanding how stable structures work and how they can be used effectively and safely.
```

docs/src/memory-manager.md
```diff
@@ -0,0 +1,37 @@
+# Memory Manager
+
+As mentioned in the previous section, each stable structure requires its own dedicated `Memory` instance.
+This is an intentional design decision that limits [the blast radius](./design-principles.md) of potential bugs, ensuring that issues only affect the specific data structure and its associated memory, not other structures.
```
dsarlis, line 4:
```suggestion
This is an intentional design decision that limits [the blast radius](./design-principles.md) of potential bugs, ensuring that issues only affect the specific stable structure and its associated memory, not other stable structures.
```

docs/src/memory-manager.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Overview
+
+The Memory Manager enables the creation of up to 255 virtual memories from a single underlying memory instance.
+When used with stable memory, this allows you to maintain up to 255 separate stable structures, each with its own isolated memory space.
+
+## Usage
```
dsarlis, line 11: In the `MemoryTrait` section we use `Usage Example` as the section's title. Should we stick to one version?

dsarlis (commented):
Should we introduce some folder structure in the docs directory that matches with the sections/sub-sections? Or is the approach of dropping these files in the same level the common approach with such books?

docs/src/memory-manager.md
ielashi: Sure. Updated it to be "usage example"

ielashi:
> Should we introduce some folder structure in the docs directory that matches with the sections/sub-sections? Or is the approach of dropping these files in the same level the common approach with such books?

Good suggestion. Done.

PR #291, opened 2025-04-29
chore: add mdbook github workflow

ielashi:
Adds generating the mdbook docs to CI. Once this commit is merged, the next step would be to setup the repo to expose this book as a github page.


docs/mdbook-admonish.css
```diff
@@ -0,0 +1,351 @@
+@charset "UTF-8";
```
dsarlis, line 1: Did you get a template from somewhere? Should there be a comment explaining where one could start if they wanted to make changes here? (Although I guess this file would be rarely if ever changed)

dsarlis (approved):

PR #292, opened 2025-04-29
feat: Add new BTreeSet functions

dragoljub-duric:
This PR adds the `is_subset`, `is_superset`, `is_disjoint`, and `symmetric_difference` functions to BTreeSet.

Reason: We already introduced `union` and `intersection` functions in the previous PR, so it makes sense to add the rest of the set operations, similar to those in the standard library.


src/btreeset.rs
```diff
@@ -567,6 +568,200 @@ where
...
+    /// set1.insert(1);
+    /// set1.insert(2);
+    /// set1.insert(3);
+    /// set2.insert(1);
+    /// set2.insert(2);
+    ///
+    /// assert!(set1.is_superset(&set2));
+    /// assert!(!set2.is_superset(&set1));
+    /// ```
+    pub fn is_superset(&self, other: &BTreeSet<K, M>) -> bool {
```
dsarlis, line 685: This does not seem particularly useful to me and I'm surprised that the Rust standard library provides it as the relationship "A is superset of B" is mathematically equivalent to "B is subset of A". Anyway if it's provided in std then it's ok for us to provide it.

src/btreeset.rs
```diff
@@ -567,6 +568,200 @@ where
...
+                        // Keys match, advance both iterators.
+                        self_next = self_iter.next();
+                        other_next = other_iter.next();
+                    }
+                    std::cmp::Ordering::Greater => {
+                        // Advance the `other` iterator if its key is smaller.
+                        other_next = other_iter.next();
+                    }
+                    std::cmp::Ordering::Less => {
+                        // If `self_key` is not found in `other`, return false.
```
dsarlis, line 649:
```suggestion
                        // `self_key` is smaller than the current smallest item of
                        // other which means that it cannot be found in `other`,
                        // so return false.
```

src/btreeset.rs
```diff
@@ -567,6 +568,200 @@ where
...
+        other: &'a BTreeSet<K, M>,
+    ) -> impl Iterator<Item = K> + 'a {
+        let mut iter_self = self.iter();
+        let mut iter_other = other.iter();
+        let mut next_self = iter_self.next();
+        let mut next_other = iter_other.next();
+
+        // Use a closure to find common elements by traversing both iterators simultaneously.
+        std::iter::from_fn(move || {
+            // Loop loop until we detect a difference or exhaust either iterator.
```
dsarlis, line 721:
```suggestion
            // Loop until we detect a difference or exhaust either iterator.
```

src/btreeset.rs
```diff
@@ -567,6 +568,200 @@ where
...
+            // This is to skip over duplicates in both iterators.
+            while next_self
+                .as_ref()
+                .zip(next_other.as_ref())
+                .is_some_and(|(a, b)| a == b)
+            {
+                next_self = iter_self.next();
+                next_other = iter_other.next();
+            }
```
dsarlis, line 730: I think we can skip doing this (and avoid the unreachable later) if we do a

```
loop {
   match (next_self.clone(), next_other.clone()) {
      ...
   }
}
```
instead below and in the case of `Equal` we advance both iterators without returning anything while in other cases returning the element we need (and advancing whichever iterator is needed).

src/btreeset.rs
dragoljub-duric: I completely agree; that was exactly what I thought.

src/btreeset.rs
dragoljub-duric: Good point, done.

dsarlis (approved):

PR #293, opened 2025-04-29
chore: Add computational complexity to doc comments of BTreeSet functions

dragoljub-duric:
This PR introduces computational complexity to the doc comments of BTreeSet functions.


src/btreeset.rs
```diff
@@ -314,6 +326,9 @@ where
     /// Returns the first key in the set. This key
```
dsarlis, line 326: Seems we did not add the complexity for `clear` above.

dsarlis (commented):
Should all of these say that `n` is the number of elements in the set or is it too obvious?

src/btreeset.rs
dragoljub-duric: Good catch. Updated, though I may be wrong. My reasoning is `fn clear_new()` takes O(n) time because clearing the allocator: The self.allocator.clear() operation deallocates all memory used by the map, which involves traversing all allocated chunks. This operation is proportional to the number of elements in the map (O(n)).

dsarlis: Yeah seems reasonable I would also make a similar argument.

PR #294, opened 2025-04-29
refactor: Refactor BTreeSet fn intersection()

dragoljub-duric:


dragoljub-duric:
@dsarlis is this what you had in mind?

dragoljub-duric:
I’m happy with either one, let me know which one you prefer.

PR #296, opened 2025-04-30
fix: Use correct feature name for canbench profiling

dsarlis:
The feature that enables specific canbench profiling of scoped code parts is named `canbench-rs` in [Cargo.toml](https://github.com/dfinity/stable-structures/blob/77ccf6cd06765cdd1ee6478c3865c27885ebf24e/Cargo.toml#L17) since a while ago. It seems that some of these were missed [when the other cases were updated](https://github.com/dfinity/stable-structures/commit/a6cdac25811d65a161f21a9df1037fe04a4f11cd).


dsarlis:
Seems there's an issue with using scopes that can be repeated in benchmarks (e.g. when you call load multiple times) that I will need to figure out.

maksymar:
> Seems there's an issue with using scopes that can be repeated in benchmarks (e.g. when you call load multiple times) that I will need to figure out.

I had that issue too. Initially I thought to create a key-value storage to record measurements for each unique scope. but it was adding too much of measurement overhead.

so I went with another approach, to record the name and timestamp data point (the fastest I came up with) and only do a post-processing to aggregating time spans for each scope when printing the stats.

see https://github.com/dfinity/stable-structures/pull/274

dsarlis:
> > Seems there's an issue with using scopes that can be repeated in benchmarks (e.g. when you call load multiple times) that I will need to figure out.
> 
> I had that issue too. Initially I thought to create a key-value storage to record measurements for each unique scope. but it was adding too much of measurement overhead.
> 
> so I went with another approach, to record the name and timestamp data point (the fastest I came up with) and only do a post-processing to aggregate time spans for each scope when printing the stats.
> 
> see #274

I was thinking that the right solution would be for canbench to allow repeated scopes. Isn't that something that others might potentially need/be restricted by? Would everyone need to implement their own custom solution like you did?

dsarlis:
> > > Seems there's an issue with using scopes that can be repeated in benchmarks (e.g. when you call load multiple times) that I will need to figure out.
> > 
> > 
> > I had that issue too. Initially I thought to create a key-value storage to record measurements for each unique scope. but it was adding too much of measurement overhead.
> > so I went with another approach, to record the name and timestamp data point (the fastest I came up with) and only do a post-processing to aggregate time spans for each scope when printing the stats.
> > see #274
> 
> I was thinking that the right solution would be for canbench to allow repeated scopes. Isn't that something that others might potentially need/be restricted by? Would everyone need to implement their own custom solution like you did?

I've got a [draft](https://github.com/dfinity/canbench/pull/94) in canbench that allows a repeated scope to cover the cases of loops or a function be called repeatedly. I think that's a better approach to deal with this instead of canbench users trying to go around it.

scripts/ci_run_benchmark.sh
maksymar, line 1: what's the diff in this file? chmod?

dsarlis: Oops, yeah. Let me undo it.

dsarlis: Fixed.

maksymar (approved):

PR #297, opened 2025-05-02
chore: Update docs for BTreeSet

dragoljub-duric:


dragoljub-duric:
I found writing this document in this way more understandable because it emphasises that we create new memory for every newly created data structure (in this case, BTreeSet).

dsarlis (approved):

PR #303, opened 2025-05-14
ci: upgrade canbench to v0.1.13

maksymar:
This PR upgrades canbench to v0.1.13 which allows to print a short benchmark summary with the link to uploaded CSV report file.


scripts/ci_run_benchmark.sh
```diff
@@ -61,27 +63,17 @@ if [ -f "$MAIN_BRANCH_RESULTS_FILE" ]; then
...
-  /\(new\)/ { print $0, "🟡"; next }
-  { print }
-  ' "$CANBENCH_OUTPUT" > "${CANBENCH_OUTPUT}.tmp" && mv "${CANBENCH_OUTPUT}.tmp" "$CANBENCH_OUTPUT"
-
-  # Add a top-level summary of detected performance changes
-  MESSAGE=""
-  grep -q "(improved " "${CANBENCH_OUTPUT}" && MESSAGE+="**🟢 Performance improvements detected! 🎉**\n"
-  grep -q "(regressed " "${CANBENCH_OUTPUT}" && MESSAGE+="**🔴 Performance regressions detected! 😱**\n"
-  echo -e "${MESSAGE:-**ℹ️ No significant performance changes detected 👍**}" >> "$COMMENT_MESSAGE_PATH"
+  CSV_RESULTS_FILE_MSG="📦 \`canbench_results_$CANBENCH_JOB_NAME.csv\` available in [artifacts](${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID})"
```
mraszyk, line 70: How about creating a single github action for that (in dfinity/canbench) instead of maintaining bash scripts across multiple repositories (I recall to have seen this already)?

mraszyk (approved):

scripts/ci_run_benchmark.sh
maksymar: This is a good idea, but I'm not experienced with CI magic, so just simplifying bash script is good enough for me for now. But this can definitely be a separate step to improve DX.

PR #308, opened 2025-05-20
feat: add optional bench_scope feature for stable-structures

maksymar:
This PR introduces an optional `bench_scope` feature for stable-structures.

When enabled, it adds `canbench_rs` scopes to selected functions.
The feature is disabled by default to avoid overhead during benchmarking.
This allows benchmarking with or without internal measurement code.

CSV report in [google sheets](https://docs.google.com/spreadsheets/d/1NVM109cRXk1DmBf_ry_4XhDGdYHx8L8nf2GtV5ruZ6E/edit?usp=sharing)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ tempfile = "3.3.0"
 test-strategy = "0.3.1"
 
 [features]
-bench_scope = [] # May increase overhead during measurement.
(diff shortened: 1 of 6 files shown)


dsarlis (approved):
I would leave a comment why these are disabled for now so that others that stumble upon this code they're not wondering.

PR #337, opened 2025-06-06
perf: read lazy size only once

maksymar:


src/btreemap/node.rs
```diff
@@ -227,18 +232,16 @@ impl<K: Storable + Ord + Clone> Node<K> {
         let size = match self.version {
             Version::V1(_) => {
                 // V1: key size is always stored in memory.
-                let size = read_u32(&reader, Address::from(offset.get()));
```
maksymar, line 230: FYI: skip reading the size again.

src/btreemap/node.rs
```diff
@@ -227,18 +232,16 @@ impl<K: Storable + Ord + Clone> Node<K> {
...
-                size
+                loaded_size
             }
             Version::V2(_) => {
-                // V2: use fixed size if available, otherwise read from memory.
+                // V2: use fixed size if available, otherwise the one loaded from memory.
                 if K::BOUND.is_fixed_size() {
                     K::BOUND.max_size()
                 } else {
-                    let size = read_u32(&reader, Address::from(offset.get()));
```
maksymar, line 239: FYI: skip reading the size again.

src/btreemap/node/v1.rs
```diff
@@ -70,15 +70,13 @@ impl<K: Storable + Ord + Clone> Node<K> {
         let mut entries = Vec::with_capacity(header.num_entries as usize);
         let mut offset = NodeHeader::size();
         for _ in 0..header.num_entries {
-            let key_offset = offset;
-            offset += U32_SIZE;
-            let key = LazyKey::by_ref(key_offset);
-            offset += Bytes::from(max_key_size);
+            let size = read_u32(memory, address + offset);
```
maksymar, line 73: V1 performance is expected to slowdown a bit, because now it starts reading the size on every load. 
But V1 is deprecated and is only expected to work correctly, users should migrate to V2.
Therefore its performance is not tracked.

michael-weigelt (approved):

PR #340, opened 2025-06-10
perf: Introduce benchmarks for the BTreeSet data structure

dragoljub-duric:
- Added a new binary target `btreeset` in `benchmarks/Cargo.toml`.
- Created `benchmarks/btreeset/src/main.rs` with benchmark functions for various `BTreeSet` operations, including:
    - `insert`
    - `remove`
    - `iter`
    - `range`
    - `union`
    - `intersection`
    - `symmetric_difference`
    - `is_subset`
    - `is_superset`
    - `is_disjoint`
- These benchmarks are generated for different key types (`u32`, `u64`, and various `Blob` sizes).


dsarlis (approved):
The title seems a bit unusual compared to our standard approach. Perhaps

> perf: Introduce benchmarks for the BTreeSet data structure

is better?

Issue #221, opened 2024-06-04
The inconsistency of `init` APIs


zhangwei983:
The `init` APIs of `StableBTreeMap` and `StableVec` returns inconsistent results:
1. `StableBTreeMap::init` returns `StableBTreeMap`
2. `StableVec::init` returns `Result<StableVec, InitError>`
Actually `MinHeap` also returns a `Result`.

Looks like `StableVec` uses the `safe_write` method which returns a `Result`. 

Should `StableBTreeMap::init` do the same thing and return a `Result`?

dsarlis:
Is there a specific concern that you have or is it mostly about consistency/uniformity of the APIs?

ielashi:
The (not very good reason) this inconsistency exists is because these structures were written by different authors. The current thinking is to update them to all not return a `Result`, since the errors returned aren't really recoverable. Would be curious to hear your opinions on this issue.

zhangwei983:
> Is there a specific concern that you have or is it mostly about consistency/uniformity of the APIs?

It's not a usage issue. For me, it's just a convention or a habit to make the high-level APIs consistent. :)

zhangwei983:
> The (not very good reason) this inconsistency exists is because these structures were written by different authors. The current thinking is to update them to all not return a `Result`, since the errors returned aren't really recoverable. Would be curious to hear your opinions on this issue.

If it's not recoverable or doesn't expect the users to handle the error, then not returning a Result is a better option IMO.

PR #351, opened 2025-06-23
refactor!: replace Result with panic in stable-structures API

maksymar:
[**Breaking change**] This PR refactors the public API of `stable-structures` to remove `Result` return types and instead panic internally on failure.

The change applies to `Vec`, `MinHeap`, `Cell`, and `Log`, affecting methods like `new`, `init`, `push`, etc.

These operations are expected to succeed under normal conditions. Returning `Result` forces callers to handle errors that are either:
- unrecoverable (e.g., memory allocation failure, memory corruption such as bad magic)
- highly unlikely edge cases (e.g., writing a `Cell` with a value larger than `u32::MAX`)

By switching to panics:
- the API becomes simpler and more ergonomic
- client code is cleaner, with no need for `.unwrap()` or `?`

This is a **breaking change** and requires updates to any downstream code that previously handled these errors.

Fixes https://github.com/dfinity/stable-structures/issues/221

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,9 @@ jobs:
       - name: Clippy
         run: cargo clippy --tests --benches -- -D clippy::all
 
+      - name: Docs
+        run: cargo doc --target wasm32-unknown-unknown
+
       - name: Test
         run: cargo test -- --color always
         env:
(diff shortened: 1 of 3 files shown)


berestovskyy (approved):
Looks good, thanks!

PR #352, opened 2025-06-23
docs: update Memory trait documentation

maksymar:
This PR documents that `Memory` trait methods panic or trap on out-of-bounds access.

### DIFF
--- a/docs/src/concepts/memory-trait.md
+++ b/docs/src/concepts/memory-trait.md
@@ -23,6 +23,13 @@ The `Memory` trait intentionally models a [WebAssembly memory instance](https://
 This design choice ensures consistency with the interface of memories available to canisters.
 It also provides future compatibility with potential multi-memory support in canisters.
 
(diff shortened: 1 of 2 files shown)


docs/src/concepts/memory-trait.md
```diff
@@ -23,6 +26,13 @@ The `Memory` trait intentionally models a [WebAssembly memory instance](https://
 This design choice ensures consistency with the interface of memories available to canisters.
 It also provides future compatibility with potential multi-memory support in canisters.
 
+## Safety contract
```
adambratschikaye, line 29: Maybe `Panics` is a better section name? Just because it is still memory safe to call these functions with out of bounds access.

adambratschikaye (approved):

docs/src/concepts/memory-trait.md
maksymar: done.

PR #354, opened 2025-06-26
docs: improve documentation of `Cell`

ielashi:


ielashi:
> I am going to do a change to public API here #351, could you please merge after that to make sure the doc is aligned with the code?

IIUC with your changes it's only a matter of removing the `expect` call from the example, right? If that's the case, I think it'd be ok to merge this, as then the compiler in your branch would just ask you to remove that `expect` method, but not strong opinions.

PR #356, opened 2025-06-30
perf: use Rc<node>

maksymar:


mraszyk (approved):
Thanks for the explanation!

PR #357, opened 2025-06-30
perf: TBD

maksymar:


src/btreemap/node/io.rs
```diff
@@ -246,32 +255,29 @@ impl<'a, M: Memory> NodeWriter<'a, M> {
...
-                (self.address + offset).get()
-            } else {
-                // Write to an overflow page, allocating it if it doesn't already exist.
-                if self.overflows.len() < page_idx {
-                    self.allocate_new_page();
-                }
-
-                (self.overflows[page_idx - 1] + offset).get()
+            let len = length.get() as usize;
+            match page_idx {
```
mraszyk, line 266: Why don't we use this match to compute the offset and then perform the write separately afterwards (to avoid code duplication)?

maksymar: I agree that this would be easier to read, and I tried it, but it turned out to be a bit slower, somewhere around +0.16% or so.

PR #374, opened 2025-07-14
test: add benchmarks with principal

maksymar:
This PR adds benchmarks for `Principal` type to `BTreeMap`.

### DIFF
--- a/benchmarks/src/random.rs
+++ b/benchmarks/src/random.rs
@@ -50,7 +50,7 @@ impl Random for u64 {
 
 impl Random for Principal {
     fn random(rng: &mut Rng) -> Self {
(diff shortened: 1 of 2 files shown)


benchmarks/src/random.rs
```diff
@@ -46,3 +47,18 @@ impl Random for u64 {
         rng.rand_u64()
     }
 }
+
+impl Random for Principal {
+    fn random(rng: &mut Rng) -> Self {
+        let size = rng.rand_u32() % Principal::MAX_LENGTH_IN_BYTES as u32;
```
berestovskyy, line 53: nit: I'm not sure if 0-sized blobs are that common in practice.

berestovskyy (approved):

Issue #166, opened 2023-12-18
BTreeMap: Explore making node loading less expensive


ielashi:
Loading is a relatively expensive operation, and improving it can lead to significant gains in performance. Caching is an obvious way to make loading a node cheaper. Maybe there are other options worth exploring too.

Related forum post:
https://forum.dfinity.org/t/is-there-any-way-to-mutate-the-value-inside-ic-stable-structures-btreemap/

PR #426, opened 2026-04-01
perf: use node cache in insert and remove operations for BTreeMap

maksymar:
This PR extends the node cache to write paths: `insert`, `remove`, `pop_first`, `pop_last`.

When descending the tree, nodes are taken from cache instead of always loading from stable memory. Unmodified parent nodes are returned to cache early (e.g. when the child is not full during insert, or when the key is not found during remove). Modified nodes go through save_node which invalidates their cache slot. Sibling nodes loaded during rebalancing bypass the cache since they are always modified and saved.

A depth parameter is threaded through recursive helpers to guide cache eviction
policy (evict deeper nodes first)

Benchmarks:
- `insert` improves ~25-30%
- `pop_first/pop_last` improves ~20%
- compound operations like `get_then_insert` improves ~25-31%

Regressions are in `first/last_key_value` (+2-5%) — read-only paths not on changed code paths, caused by write operations filling pinned depth 0-1 cache slots that read traversals also need.

Pop improvements are smaller than insert because single-pass pop (#417) already eliminates redundant traversals — cache only helps by reusing root and near-root nodes across consecutive calls.

Related to https://github.com/dfinity/stable-structures/issues/166.

chart_by_category

Comparing to base commit 04fe601f805a8c69f44a845255a5c493900db820 before all cache optimizations.
chart_by_category

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@
 **/*~
 
 tmp/
+
+# Claude Code local settings
+.claude/settings.local.json
--- a/benchmarks/btreemap/canbench_results.yml
+++ b/benchmarks/btreemap/canbench_results.yml
@@ -9,63 +9,63 @@ benches:
   btreemap_v2_contains_blob_256_128:
     total:
       calls: 1
-      instructions: 899093643
+      instructions: 899462217
       heap_increase: 0
       stable_memory_increase: 0
     scopes: {}
   btreemap_v2_contains_blob_32_0:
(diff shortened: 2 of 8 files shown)


maksymar:
after split node optimization
chart_by_category

sasa-tomic (approved):
Great work, LGTM!

mraszyk (approved):

PR #430, opened 2026-04-03
perf: lazy sibling loading in BTreeMap::remove

maksymar:
This PR implements lazy sibling loading in `BTreeMap::remove` to avoid redundant node reads.

In remove Case 3 (child at minimum), both left and right siblings were always loaded from stable memory even though only one may be needed. This change loads the left sibling first and skips loading the right sibling entirely when the left can donate an entry. Saves one node load in ~50% of rebalancing cases.

Related to https://github.com/dfinity/stable-structures/issues/166.

### DIFF
--- a/benchmarks/btreemap/canbench_results.yml
+++ b/benchmarks/btreemap/canbench_results.yml
@@ -51,21 +51,21 @@ benches:
   btreemap_v2_contains_then_remove_blob_32_128:
     total:
       calls: 1
-      instructions: 916746069
+      instructions: 916748064
       heap_increase: 0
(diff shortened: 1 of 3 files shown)


maksymar:
vs_main_by_category

cache_effect_by_category

vs_old_by_category

maksymar:
Closing — benchmarked in isolation against current main (cache=16): ~1.8% improvement on remove, no effect on other operations. Not worth the added code duplication and complexity.

If we revisit this later and the benefit becomes more significant (e.g. with a different cache strategy), we can reopen.

PR #434, opened 2026-04-21
feat(storable): implement Storable for unbounded 2-tuples

lpahlavi:
## Summary

- Replaces the two `todo!()` panics in the `(A, B)` `Storable` impl (`into_bytes_inner_2` and `from_bytes`) with a working implementation for unbounded tuples.
- The bounded 2-tuple wire format is **unchanged** — no backwards-compatibility break.

## Wire format for unbounded `(A, B)`

Documented in a comment inside the impl block, mirroring the 3-tuple style:

```
If A is fixed-size:
  <a_bytes> <b_bytes>
Otherwise:
  <size_lengths (1B)> <size_a (1-4B)> <a_bytes> <b_bytes>
```

`size_lengths` encodes (in 2 bits) how many bytes were used to represent A's length — identical to the encoding already used by the 3-tuple path. B is always the last element so its length is inferred from the remaining bytes and never stored explicitly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/storable/tuples.rs
+++ b/src/storable/tuples.rs
@@ -8,6 +8,11 @@ where
     A: Storable,
     B: Storable,
 {
+    // Tuple (A, B) serialization:
+    //   If A has fixed size:
+    //     <a_bytes> <b_bytes>
+    //   Otherwise:
+    //     <size_lengths (1B)> <size_a (1-4B)> <a_bytes> <b_bytes>
     fn to_bytes(&self) -> Cow<'_, [u8]> {
         Cow::Owned(into_bytes_inner_2(Self::BOUND, &self.0, &self.1))
     }
@@ -42,7 +47,24 @@ where
                 let b = B::from_bytes(Cow::Borrowed(&bytes[a_max_size..a_max_size + b_len]));
                 (a, b)
             }
-            _ => todo!("Deserializing tuples with unbounded types is not yet supported."),
+            Bound::Unbounded => {
+                let mut offset = 0;
+                let mut size_length_a = None;
+
+                if !A::BOUND.is_fixed_size() {
+                    let lengths = decode_size_lengths(bytes[0], 1);
+                    offset += 1;
+                    size_length_a = Some(lengths[0]);
+                }
+
+                let (a, read) = decode_tuple_element::<A>(&bytes[offset..], size_length_a, false);
+                offset += read;
+                let (b, read) = decode_tuple_element::<B>(&bytes[offset..], None, true);
+                offset += read;
+
+                debug_assert_eq!(offset, bytes.len());
+                (a, b)
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/storable/tuples.rs
```diff
@@ -8,6 +8,13 @@ where
     A: Storable,
     B: Storable,
 {
+    // Tuple (A, B) serialization:
+    //   Bounded (A and B are both bounded):
+    //     <a_bytes (max_a B, zero-padded)> <b_bytes (max_b B, zero-padded)> <size_a (0-4B)> <size_b (0-4B)>
```
alin-at-dfinity: I don't understand what this is trying to say. Shouldn't it just be `<a_bytes> <b_bytes>`?

alin-at-dfinity (approved):

src/storable/tuples.rs
lpahlavi: Good catch! That comment was... not great. I reformulated it to be clearer and mirror the similar note we have below for 3-tuples.

maksymar:
(for posterity) 

I just noticed this PR and I would like to provide some context.

Unbounded 2-tuple was originally left unimplemented because [EXC-1550](https://dfinity.atlassian.net/browse/EXC-1550) explored a smooth `Bounded`-to-`Unbounded` transition and found it infeasible. The legacy bounded layout (`<a(padded)> <b(padded)> <size_a> <size_b>`) stores sizes at offsets derived from `a_max`/`b_max`, which aren't recoverable without the `Bound` metadata. The ticket was descoped.

The concern this PR introduces: the 2-tuple now carries two completely different wire formats, so a user who flips a type's `BOUND` from `Bounded` to `Unbounded` will silently mis-decode existing data. The 3-tuple doesn't have this problem, it uses one unified wire format for both variants, so as long as `is_fixed_size()` is preserved, a flip is a no-op at the tuple's wire level.

Worth re-evaluating whether shipping this is better than the previous `todo!()`, which at least failed loudly.

If we keep it I'd want at minimum:
- doc warning on the tuple `Storable` impl that flipping type's bound is a silent wire-format break for 2-tuples
- document the bounded format in the 2-tuple impl comment alongside the unbounded one
- add `(fixed, unbounded)` and maybe other combinations of roundtrip tests
- and long-term, if we keep it now, then if we decide to unify the 2-tuple later it'll be a breaking change

@lpahlavi what do you think on whether we should keep it and face a silent-breakage risk or revert it?


[EXC-1550]: https://dfinity.atlassian.net/browse/EXC-1550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #437, opened 2026-04-22
feat: support unbounded 2-tuples in `Storable`

ielashi:
Previously encoding `(A, B)` tuple with Storable was only possible if A and B are `Bounded`. This commit extends that support to allow A and B to be `Unbounded`.


maksymar:
hi Islam, I just recently asked Louis to revert the same PR here: https://github.com/dfinity/stable-structures/pull/436.
not because of the implementation, but because of the correct release procedure, just to make sure we are not releasing a silent footgun.

PR #441, opened 2026-07-28
perf(btreemap): batch the per-field node reads and writes

hpeebles:
Loading a node performed one memory read per field — every child address, key size, key, and value size separately — dozens of reads per node, and saving mirrored that with one write per field. A memory access costs a fixed amount per call plus a fee per byte, so this PR batches both directions while keeping every path that can't be batched profitably on its existing per-field behavior.

### Load

The load now fetches the front of the initial page in one read and parses the fields out of the buffer, with any field beyond the read falling back to an individual read. The read is sized so that it only ever buys bytes the load will use:

- For key types bounded by ≤ 64 bytes, the header pins down the extent of the children and key areas, and that estimate is all that is fetched.
- For larger or unbounded key bounds, the interleaved layout offers no extent worth prefetching — the size fields the walk needs sit between payloads it skips — so those types read per-field exactly as before. (An earlier draft that read a fixed 1 KiB for them regressed the large-value benchmarks by up to 8%.)

The previous `CHILDREN_BATCH_READ_THRESHOLD = 4` heuristic ("individual reads are faster for ≤ 4 children") is gone: re-measured under the new read pattern, it no longer paid for itself — the sweep below shows no regression on any internal-node-heavy benchmark.

### Save

Fields are serialized into a buffer flushed in as few writes as possible, except that slices of ≥ 48 bytes are written directly, uncopied, flushing whatever is buffered around them. The threshold is measured, and lower than intuition suggests: copying into the buffer costs a few instructions per byte, while a separate write costs a fixed amount for the call and its page mapping, putting the crossover around forty bytes. (Drafts with 128- and 512-byte thresholds both measured thousands of instructions per save slower on nodes with ~128-byte keys and values.)

The save no longer writes the overflow-address field at bytes [7, 15): the `NodeWriter` owns it, updating the chain pointer in place as overflow pages are allocated and released, and `finish()` writes the `NULL` terminator into it whenever the node ends up with no overflow pages — so even a fresh node at a reused address cannot retain a stale pointer. (The old explicit write was redundant with that terminator.)

A new test covers the one newly reachable trap: a node placed so close to the end of memory that the batched-read estimate extends past the last byte ever written; the read clamps rather than traps.

### Benchmarks

Full `canbench` sweep against `main` (canbench 0.2.0, all 252 benchmarks):

```
counts:   [total 252 | regressed 0 | improved 147 | new 0 | unchanged 105]
change %: [max +1.84% | p75 -0.01% | median -3.87% | p25 -11.92% | min -23.48%]
```

**Zero regressions.** The single largest positive movement (+1.84%, `remove_vec_64_128`) is on a code path this PR does not touch (unbounded keys take the unchanged per-field route) and is within canbench's noise threshold.

Top improvements:

| benchmark | Δ instructions |
|---|---|
| `mem_manager_remove_u64_u64` | −23.48% |
| `mem_manager_insert_u64_u64` | −22.11% |
| `mem_manager_get_u64_u64` | −19.87% |
| `get_blob8_u64` | −18.48% |
| `get_zipf_100k_u64_u64_nocache` | −17.58% |
| `insert_blob8_u64` | −16.91% |
| `remove_blob8_u64` | −16.70% |
| `mem_manager_get_u64_blob512` | −16.08% |
| `pop_last_blob_32_0` | −16.01% |
| `scan_iter_rev_1k_0b` | −15.59% |
| `contains_blob_8_128` | −15.53% |
| `get_blob_4_128` | −15.27% |
| `remove_100k_u64_u64_nocache` | −14.70% |

Memory-manager maps gain the most (−16 to −23%), followed by maps with small keys and values (−10 to −19%); sparse-node scans and range queries land around −14 to −16%. One footnote: `insert_vec_8_128` shows a one-time +1 page (64 KiB) heap high-water increase from the save buffer — a peak-allocation effect, not a per-operation cost.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/btreemap/node/io.rs
+++ b/src/btreemap/node/io.rs
@@ -281,22 +281,6 @@ impl<'a, M: Memory> NodeWriter<'a, M> {
         }
     }
 
-    pub fn write_u32(&mut self, offset: Address, val: u32) {
-        self.write(offset, &val.to_le_bytes());
-    }
-
-    pub fn write_u64(&mut self, offset: Address, val: u64) {
-        self.write(offset, &val.to_le_bytes());
-    }
-
-    pub fn write_struct<T>(&mut self, t: &T, addr: Address) {
-        let slice = unsafe {
-            core::slice::from_raw_parts(t as *const _ as *const u8, core::mem::size_of::<T>())
-        };
-
-        self.write(addr, slice)
-    }
-
     // Allocates a new page and appends it to the node's overflows.
     fn allocate_new_page(&mut self) {
         let new_page = self.allocator.allocate();
--- a/src/btreemap/node/tests.rs
+++ b/src/btreemap/node/tests.rs
@@ -309,3 +309,32 @@ fn can_call_node_value_multiple_times_on_same_index() {
     let value2 = node.value(0, &mem);
     assert_eq!(value1, value2);
 }
+
+#[test]
+fn loading_near_the_end_of_memory_clamps_the_batched_read() {
+    use crate::storable::Blob;
+
+    // A bounded-key node placed so that the load's batched-read estimate extends past
+    // the last byte of memory: the read must clamp to the memory's end rather than
+    // trap. (The node's own bytes always fit — the save wrote them — it is only the
+    // estimate, an upper bound based on the key bound, that can overshoot.)
+    let mem = make_memory();
+    let page_size = PageSize::Value(128);
+    // The allocator's header write grows the memory to exactly one Wasm page.
+    let mut allocator = Allocator::new(mem.clone(), Address::from(0), Bytes::from(128u64));
+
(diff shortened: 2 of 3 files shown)


hpeebles:
I've just been told that there are plans to make stable read/writes as fast (or thereabouts) as heap read/writes. Which would make this PR unnecessary.
So I'll close this, rather than add complexity for a temporary gain.

hpeebles:
It turns out that even if stable read/writes are just as fast as to the heap, the changes in this PR would still actually result in roughly the same proportional improvements, since with these changes applied there are far fewer read/writes through the `NodeReader`/`NodeWriter`, each of which incurs a small cost.
So I think it is worth re-opening this PR in the future if performance becomes a more critical issue.

